Issue #4, opened 2022-09-06
Setup theme package


richbachman:
Work with design team to set up the Theme package. The theme package is the foundation of the design system. It will control all color, sizing, etc for the components. The package will export a [Stitches theme](https://stitches.dev/docs/tokens), along with any globals styles and Stitches functions. The components package will use this theme package to style components.

**AC:**
- Export global styles
- Export a default theme
- Includes all configs for NPM publishing (publish will be handled in a future ticket)

PR #5, opened 2022-09-06
feat(theme): setup theme package

richbachman:
## Description of the change

This packages sets up the initial theme package based off this [Figma WIP](https://www.figma.com/file/uDEjU77uIrQ5T1csn8IvIJ/%5BWIP%5D-Pluto-Tokens?node-id=0%3A1) file.

The Stitches theme contains the tokens for future use within components. This also contains a global stylesheet function to be used in storybook and other consuming applications.

This PR also includes some minor package.json changes, contributing doc changes, and lint-stage changes.

Closes: https://github.com/Localitos/pluto/issues/4

## Testing the change

Because the theme is basically a `json` file of tokens, the best option to test is via snapshot testing.

## Type of change

### Development

### Code review

- [ ] Issue from Linear has a link to this pull request


packages/theme/src/theme/default.ts
```diff
@@ -1,149 +1,146 @@
...
+      },
+      fontSizes: {
+        fontSize10: "0.75rem", // 12px
+        fontSize20: "0.875rem", // 14px
+        fontSize30: "1rem", // 16px
+        fontSize40: "1.25rem", // 20px
+        fontSize50: "1.5rem", // 24px
+        fontSize60: "2rem", // 32px
+        fontSize70: "2.5rem", // 40px
+        fontSize80: "3.375rem", // 54px
```
RohitRajendran, line 67: Will Figma allow showing the font size as `fontSize10` when inspecting a mock up? If not, I think it gets a bit annoying to translate pixels/rems to the right font size name.

richbachman: **Short answer:** Yes.

**Slightly longer answer:** It's going to take some setup on the Figma side to accomplish that. As the design team works through creating symbols and components with the new tokens, there are some plugin options that will display that info in the inspect panel. I'll be working with them to figure out the best option.

**Side note:** I totally understand that using a scale token naming scheme is difficult to translate at the beginning. As the system gets used more, that part gets easier. The naming allows us to scale as well. Using a scale naming convention disconnects the token name from the actual value, allowing the value to be anything. So `fontSize30` can be `16px` now, but in the future it could be `18px`. In order to implement that change, we'd only have to change the token value and not the token uses in each component.

RohitRajendran: Awesome, makes sense!

packages/theme/src/styles/global.ts
```diff
@@ -0,0 +1,26 @@
+import { globalCss } from "@stitches/react";
+
+export const globalStyles = globalCss({
```
RohitRajendran, line 3: Just wanted to check if any of this is different than what's currently in the application? Will we need to fix any stylings in the application because it's not expecting any of these global styles?

RohitRajendran (approved):
Had one question but looks good!

packages/theme/src/styles/global.ts
richbachman: Good question. This is a slimmed down version of whats in the current [index.scss file](https://github.com/Localitos/Backend/blob/main/client/src/index.scss). I actually removed the margin and padding reset since thats a bit heavy handed and would cause some issues. The `fontSize: "100%",` is a safe addition.

PR #31, opened 2022-09-21
feat: setup chromatic to Pluto

apvale:
## Description of the change

This PR setups Chromatic to Pluto:

- [x] Set up Github action workflow for Chromatic.

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/.github/workflows/visual_regression_tests.yml
+++ b/.github/workflows/visual_regression_tests.yml
@@ -1,0 +1,56 @@
+name: "Visual Tests"
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  push:
+    paths:
+      - "packages/**"
+    branches:
+      - main
+
+jobs:
+  install-cache:
+    name: Install and cache dependencies
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
+    steps:
+      - name: Checkout Commit
+        uses: actions/checkout@v2
+      - name: Cache yarn dependencies
+        uses: actions/cache@v3
+        id: yarn-cache
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-yarn-v1-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-yarn-v1
+      - name: Install dependencies if cache invalid
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn
+
+  visual-tests:
+    name: Run visual tests with Chromatic
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.draft == false
+    needs: install-cache
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Required to retrieve git history
+      - name: Restore yarn dependencies
+        uses: actions/cache@v3
+        id: yarn-cache
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-yarn-v1-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-yarn-v1
+
+      - name: Build packages
+        run: yarn build
+
+      - name: Publish to Chromatic
+        uses: chromaui/action@v1
+        with:
+          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          autoAcceptChanges: "main"
+          exitOnceUploaded: true
--- a/package.json
+++ b/package.json
@@ -1,7 +1,10 @@
 {
   "name": "pluto",
   "version": "1.0.0",
-  "repository": "git@github.com:Localitos/pluto.git",
+  "repository": {
+    "type": "git",
+    "url": "git+ssh://git@github.com/Localitos/pluto.git"
+  },
   "license": "MIT",
   "private": true,
   "scripts": {
@@ -68,5 +71,11 @@
   },
   "browserslist": [
     "extends @localyze-pluto/browserslist-config"
-  ]
+  ],
+  "bugs": {
+    "url": "https://github.com/Localitos/pluto/issues"
+  },
+  "readme": "ERROR: No README data found!",
+  "homepage": "https://github.com/Localitos/pluto#readme",
+  "_id": "pluto@1.0.0"
 }
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@
 # Jest
 .jest-coverage
 .jest-cache
+
+# Design system storybook build
+/docs
--- a/cspell.json
+++ b/cspell.json
@@ -15,7 +15,8 @@
     "testid",
     "keyof",
     "toshimaru",
-    "webp"
+    "webp",
+    "chromaui"
   ],
   "flagWords": ["hte"],
   "allowCompoundWords": true,
--- a/packages/components/src/primitives/Box/BaseBox.tsx
+++ b/packages/components/src/primitives/Box/BaseBox.tsx
@@ -14,13 +14,11 @@ export interface BaseBoxProps
 
 /** The BaseBox is the basis for Box. It handles the ref forwarding and other baseline props. */
 const BaseBox = React.forwardRef<HTMLDivElement, BaseBoxProps>(
-  ({ children, decorative, ...props }, ref) => {
-    return (
-      <div aria-hidden={decorative ? true : undefined} ref={ref} {...props}>
-        {children}
-      </div>
-    );
-  }
+  ({ children, decorative, ...props }, ref) => (
+    <div aria-hidden={decorative ? true : undefined} ref={ref} {...props}>
+      {children}
+    </div>
+  )
 );
 
 BaseBox.displayName = "BaseBox";
--- a/packages/components/src/primitives/Box/BaseBox.tsx
+++ b/packages/components/src/primitives/Box/BaseBox.tsx
@@ -14,11 +14,13 @@ export interface BaseBoxProps
 
 /** The BaseBox is the basis for Box. It handles the ref forwarding and other baseline props. */
 const BaseBox = React.forwardRef<HTMLDivElement, BaseBoxProps>(
-  ({ children, decorative, ...props }, ref) => (
-    <div aria-hidden={decorative ? true : undefined} ref={ref} {...props}>
-      {children}
-    </div>
-  )
+  ({ children, decorative, ...props }, ref) => {
+    return (
+      <div aria-hidden={decorative ? true : undefined} ref={ref} {...props}>
+        {children}
+      </div>
+    );
+  }
 );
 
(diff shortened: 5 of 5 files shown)


.github/workflows/on_release_publish.yml
```diff
@@ -39,6 +39,18 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.TOKEN_BOT }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      # Chromatic
+      - name: Build Storybook
+        run: yarn build-storybook
+
+      - name: Run Chromatic
```
padulafacundo, line 46: I wonder if this should be its own job. Do we need all the previous steps to run chromatic?

.github/workflows/on_release_publish.yml
```diff
@@ -39,6 +39,18 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.TOKEN_BOT }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      # Chromatic
+      - name: Build Storybook
+        run: yarn build-storybook
```
padulafacundo, line 44: Can't we pass `buildScriptName` to the chromatic action?

> ```buildScriptName: 'The npm script that builds your Storybook [build-storybook]'```

apvale: Why do you think we need this? 🤔

padulafacundo: We don't _need_ it. But from the docs I understand that we could pass `buildScriptName: build-storybook` to the chromatic action and it will take care of building storybook, no? And we could remove this previous step ☝️

apvale: I updated the yml removing the unnecessary steps. We don't need the `buildScriptName` because our build command is the default one (`build-storybook`). 🙏🏻 

Can you check it again, please?

.github/workflows/on_release_publish.yml
```diff
@@ -10,6 +10,24 @@ on:
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
+  # Run visual tests with Chromatic
```
padulafacundo, line 13: Is this repeated in `ui-tests.yml` ?

apvale: Yes, it is.

apvale: I want to run the visual tests when opening a PR and also when pushing to main. Maybe I can update this: 
```
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]

```

?

apvale: Updated to this:

```
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
  push:
    branches:
      - main
```

.github/workflows/ui-tests.yml
```diff
@@ -0,0 +1,54 @@
+name: "Visual Tests"
+
+on:
```
padulafacundo, line 3: Maybe we should have this only for changes in the packages? We shouldn't run visula tests if we update the readme or change something in the build

richbachman: Yeah. We should on run this if there are package changes. Let's also rename this file to `on_pull_requests_vrt.yml`.

package.json
```diff
@@ -19,7 +22,8 @@
     "test": "jest",
     "test:ci": "jest --ci",
     "test:coverage": "jest --runInBand --coverage",
-    "type-check": "turbo run tsc"
+    "type-check": "turbo run tsc",
+    "chromatic": "npx chromatic --project-token=90f3c5364d09"
```
richbachman, line 26: We don't need to install Chromatic in the repo. We can just rely on the GH action to do handle running Chromatic.

.github/workflows/ui-tests.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  push:
+    paths:
+      - "packages/**"
+    branches:
+      - main
```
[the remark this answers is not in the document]
apvale: It also runs in the current branch because of the `synchronize` type of the PR.

.github/workflows/ui-tests.yml
apvale: This will run for `on_pull_requests` and also for pushing to main. How would this file name be better for these two actions?

richbachman:
For this one, we need to set up two Github workflows:
- Run Chromatic on every PR that changes a package
- Run Chromatic on main when a package has changed

The package PR one should install the packages, build the packages, and then run Chromatic. The merge to main one should to the same. It is a double check to make sure something didn't break on main. You may also need to add the following to the action in its `with` section.

```
autoAcceptChanges: "main"
exitOnceUploaded: true
```

.github/workflows/ui-tests.yml
apvale: I added this to this action: 
```
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
  push:
    paths:
      - "packages/**"
    branches:
      - main
```

But how can I add a check in the current branch? I mean, we want to check this for the main and also for the branch that belongs to the PR.

richbachman: Can you do:
```
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
    paths:
      - "packages/**"
  push:
    paths:
      - "packages/**"
    branches:
      - main
```

apvale: I did https://github.com/Localitos/pluto/pull/31/files#diff-164c684a80d38b999b17c7abe8d6485a68155c70969e986bdd1e53386b42eb1eR6 
And it worked. But I'm not sure if this is good. Sometimes the publish to Chromatic might not work and our next push wouldn't be a change in the packages folder, for example, so that it wouldn't trigger the job. 🤷🏻‍♀️ 

I tried this case on my machine and it did trigger the Visual tests job, but I'm still not sure if this will always work as we expect x)

apvale: Yeah, exactly. I updated the packages and ran this build https://www.chromatic.com/build?appId=632b1baf1257171b6a84f370&number=25
But then I updated it back and committed. But instead of waiting to finish the jobs, I decided to merge with main and then committed again. It didn't generate a new build with the update that I did in the packages. 

I prefer doing it this way:

```
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
  push:
    paths:
      - "packages/**"
    branches:
      - main
```

Running in the `packages` folder only for branch `main`. And I'm also not sure if this will work correctly using the same action for two different situations. I'm also open to trying and seeing if it works smoothly. 🤞🏻

package.json
```diff
@@ -70,5 +73,14 @@
   },
   "browserslist": [
     "extends @localyze-pluto/browserslist-config"
-  ]
+  ],
+  "dependencies": {
+    "chromatic": "^6.9.0"
+  },
```
richbachman, line 79: We can remove this dep.

.github/workflows/ui-tests.yml
richbachman: Yeah. Let's try this method out, and we can always adjust if necessary.

.github/workflows/visual_regression_tests.yml
```diff
@@ -0,0 +1,58 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Commit
+        uses: actions/checkout@v2
+      - name: Cache yarn dependencies and cypress
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: |
+            node_modules
```
padulafacundo, line 24: nit: Could this be just one line?

.github/workflows/visual_regression_tests.yml
```diff
@@ -0,0 +1,58 @@
...
+          key: ${{ runner.os }}-yarn-v1-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-v1
```
padulafacundo, line 27: Same for this

.github/workflows/visual_regression_tests.yml
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  install-cache:
+    name: Install and cache dependencies
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Commit
+        uses: actions/checkout@v2
+      - name: Cache yarn dependencies and cypress
+        uses: actions/cache@v2
```
padulafacundo, line 20: There's a v3 for the cache action. Do you know if we could use that?

.github/workflows/visual_regression_tests.yml
```diff
@@ -0,0 +1,58 @@
+name: "Visual Tests"
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
padulafacundo, line 5: Do we need `ready_for_review` ? Or `reopened`? I mean, do we need to trigger the tests if nothing has changed (no diff) ?

.github/workflows/visual_regression_tests.yml
```diff
@@ -0,0 +1,58 @@
...
+    needs: install-cache
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # Required to retrieve git history
+      - name: Restore yarn dependencies
+        uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: |
```
padulafacundo, line 44: Same about one line

package.json
apvale: Removed. 🙏🏻

.github/workflows/visual_regression_tests.yml
apvale: Updated to v3.

.github/workflows/visual_regression_tests.yml
apvale: I'm not sure. Maybe not. I just used this from an example. Why do you think we don't need it?

padulafacundo: Well, I think we should only trigger vrt when the result of the tests can change. Said differently, is there a scenario where a test will change its output (pass or fail) after marking a PR as ready for review? Or after closing it and opening?
No changes are introduced to any file in those scenarios, so I suppose the test will return the same result? So we will run it twice when we actually don't need to

padulafacundo: Ok, just saw you added a conditional to only run tests if it's not a draft. Depending on how we work on it I think there might be value on running tests for draft PRs, but I don't feel too strong about this. Let's go with this and see how it plays out

padulafacundo (approved):
Looks good to me 👍

nyan07 (approved):

PR #93, opened 2022-10-14
feat(components): add avatar

nyan07:
## Description of the change
This PR adds the avatar component with its six different sizes.

image

## Testing the change
Run storybook

## Type of change

### Development

### Code review


packages/components/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,172 @@
...
+  w: string;
+  h: string;
+  margin: SystemProp<keyof Theme["space"], Theme>;
+} => {
+  switch (size) {
+    case 6: {
+      return {
+        w: "24px",
+        h: "24px",
+        margin: "space30", // Should be 10px - no token provided
```
nyan07, line 31: @richbachman, the design team used a space around the avatar for which we don't have a token. But the margin property only accepts tokens. Which do you think should be our approach here?

packages/components/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,172 @@
...
+    }
+    case 10:
+    case 12: {
+      return {
+        fontSize: "fontSize30",
+      };
+    }
+    // Case 12: {
+    //   Return {
+    //     FontSize: "fontSize30", // Should be 18px - no token provided
```
nyan07, line 96: @richbachman same regarding the token here.

packages/theme/src/theme/default.ts
```diff
@@ -52,6 +52,7 @@ export const theme = {
     colorIconWeak: "#64748B",
     colorIconStrong: "#334155",
     colorIconStronger: "#0F172A",
+    colorAvatarInitials: "#4A63FC",
```
nyan07, line 55: @richbachman the idea is to add the colors here or just keep the ones we already have? Maybe this one we can use this directly on the component because I don't believe that it will be used in another place.

packages/components/src/components/Avatar/Avatar.tsx
richbachman: Let's check in with design and see if the 8px or 12px token will work. I had to do that with the Button component. I don't think they've had chance to go back and edit some of the FIgma files since they removed the 10px spacing.

packages/components/src/components/Avatar/Avatar.tsx
richbachman: Same here. Lets just question design on it.

packages/theme/src/theme/default.ts
richbachman: We'll be adding to these as components get built out. I'm cool with having a component specific token color here. Since its most likely the only place it will be used. 👍

apvale (approved):

PR #123, opened 2022-10-24
fix(components): fix Callout types and remove role

nyan07:
## Description of the change
Fix Callout types and remove role alert. 

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Callout/Callout.tsx
+++ b/packages/components/src/components/Callout/Callout.tsx
@@ -5,7 +5,8 @@ import { Icon, IconProps } from "../Icon";
 
 type CalloutVariantOptions = "error" | "information" | "success" | "warning";
 
-type CalloutProps = {
+interface CalloutProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, "color"> {
   /** The content of the callout. It can be text or valid text related HTML, i.e. anchor and strong elements. */
   children: React.ReactNode;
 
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Callout/Callout.tsx
```diff
@@ -126,7 +127,6 @@ const Callout = React.forwardRef<HTMLDivElement, CalloutProps>(
         gap="space40"
         padding="space50"
         ref={ref}
-        role="alert"
```
apvale, line 129: Why remove this?

packages/components/src/components/Callout/Callout.tsx
```diff
@@ -14,7 +15,7 @@ type CalloutProps = {
 
   /** Show or hide the callout background */
   withoutBackground?: boolean;
-};
+}
```
apvale, line 18: Shouldn't we keep the`;` ?  What are the defaults? 😭

packages/components/src/components/Callout/Callout.tsx
nyan07: Because a callout is not an alert. The alert is a role we use when something appears on the screen, and we need the user to immediately look into the content that just appeared.

packages/components/src/components/Callout/Callout.tsx
nyan07: This line should be whithout `;` because it closes the type. We should have `;` in each line inside it.

PR #126, opened 2022-10-24
feat(components): add badge

blumaa:
## Description of the change

This is a PR to add the Badge component

Screenshot 2022-10-26 at 16 33 51
Screenshot 2022-10-26 at 16 34 03
Screenshot 2022-10-26 at 16 34 14

## Testing the change

- [x] Write your testing instructions here

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Badge/Badge.stories.tsx
+++ b/packages/components/src/components/Badge/Badge.stories.tsx
@@ -1,0 +1,60 @@
+import type { ComponentMeta, ComponentStory } from "@storybook/react";
+import React from "react";
+import { Badge } from "./Badge";
+
+export default {
+  component: Badge,
+  title: "Components/Badge",
+} as ComponentMeta<typeof Badge>;
+
+const Template: ComponentStory<typeof Badge> = (args) => <Badge {...args} />;
+
+export const Default = Template.bind({});
+Default.args = {
+  children: "Primary badge",
+};
+
+export const DefaultBadgeWithIcon = Template.bind({});
+DefaultBadgeWithIcon.args = {
+  children: "Blue badge",
+  icon: "EyeIcon",
+};
+
+export const Gray = Template.bind({});
+Gray.args = {
+  children: "Gray badge",
+  color: "gray",
+};
+
+export const Green = Template.bind({});
+Green.args = {
+  children: "Green badge",
+  color: "green",
+};
+
+export const Yellow = Template.bind({});
+Yellow.args = {
+  children: "Yellow badge",
+  color: "yellow",
+};
+
+export const Large = Template.bind({});
+Large.args = {
+  children: "Large badge",
+  size: "large",
+};
+
+export const LargeBadgeWithIcon = Template.bind({});
+LargeBadgeWithIcon.args = {
+  children: "Large badge with Icon",
+  size: "large",
+  icon: "CheckCircleIcon",
+};
+
+export const LargeYellowBadgeWithIcon = Template.bind({});
+LargeYellowBadgeWithIcon.args = {
+  children: "Large badge with icon",
+  icon: "ArrowDownCircleIcon",
+  size: "large",
+  color: "yellow",
+};
--- a/packages/components/src/components/Badge/Badge.tsx
+++ b/packages/components/src/components/Badge/Badge.tsx
@@ -1,0 +1,116 @@
+import type { SystemProp, Theme } from "@xstyled/styled-components";
+import React from "react";
+import * as HeroOutlineIcons from "@heroicons/react/24/outline";
+import { Box } from "../../primitives/Box";
+import { Icon } from "../Icon";
+
+type IconNames = keyof typeof HeroOutlineIcons;
+
+type BadgeSizeOptions = "large" | "small";
+type BadgeColorOptions = "blue" | "gray" | "green" | "yellow";
+
+export interface BadgeProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, "color"> {
+  /** The contents of the badge. Can be text. */
+  children: NonNullable<React.ReactNode>;
+  /** Icon to be added on the left of the content. */
+  icon?: IconNames;
+  /** Sets the size of the badge. */
+  size?: BadgeSizeOptions;
+  /** Sets the color of the badge. */
+  color?: BadgeColorOptions;
+}
+
+const getColorProps = (
+  color: BadgeColorOptions
+): {
+  backgroundColor?: SystemProp<keyof Theme["colors"], Theme>;
+  color?: SystemProp<keyof Theme["colors"], Theme>;
+} => {
+  switch (color) {
+    case "gray": {
+      return {
+        backgroundColor: "colorBackgroundWeak",
+        color: "colorTextStronger",
+      };
+    }
+    case "green": {
+      return {
+        backgroundColor: "colorBackgroundSuccess",
+        color: "colorTextSuccess",
+      };
+    }
+    case "yellow": {
+      return {
+        backgroundColor: "colorBackgroundWarning",
+        color: "colorTextWarning",
+      };
+    }
+    default: {
+      return {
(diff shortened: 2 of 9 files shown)


nyan07:
You need to export it in this file `packages/components/src/index.ts`

blumaa:
@nyan07 It's not enough to have `export * from "./Badge";` in index.ts? This is how the other components are being exported.

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+const getIcon = (iconName: IconNames, size: BadgeSizeOptions) => {
+  const iconProps = (size: BadgeSizeOptions) =>
+    size === "small" ? "sizeIcon30" : "sizeIcon40";
+
+  return <Icon decorative icon={iconName} size={iconProps(size)} />;
+};
+
+/** This is a badge */
```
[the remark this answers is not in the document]
blumaa: Yes, but what do you suggest? I'm not entirely sure what a badge is used for! "A badge is used in lists to emphasise statuses... "?

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,106 @@
...
+  | "primary"
+  | "success"
+  | "warning";
+
+export interface BadgeProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, "color"> {
+  /** The contents of the badge. Can be text. */
+  children: NonNullable<React.ReactNode>;
+  /** Sets the badge to be disabled. This is still being discussed with design whether we need the disabled prop or not. */
+  isDisabled?: boolean;
```
nyan07, line 22: I would prefer `disabled` because we are not starting boolean props with `is` in other components.

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,106 @@
...
+      };
+    }
+  }
+};
+
+const getIcon = (iconName: IconNames, size: BadgeSizeOptions) => {
+  const iconProps = (size: BadgeSizeOptions) =>
+    size === "small" ? "sizeIcon30" : "sizeIcon40";
+
+  return <Icon decorative icon={iconName} size={iconProps(size)} />;
```
nyan07, line 75: Could an icon be on the right side? :thinking:  Or is it always on the left (maybe it would be better to check it with the design team)?

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,106 @@
...
+  ({ children, size = "small", variant = "primary", icon, ...props }, ref) => {
+    return (
+      <Box.div
+        alignItems="center"
+        borderRadius="borderRadiusPill"
+        display="inline-flex"
+        fontSize={size === "large" ? "fontSize20" : "fontSize10"}
+        fontWeight="fontWeightBold"
+        gap="space20"
+        paddingBottom={size === "large" ? "space30" : "space20"}
```
nyan07, line 89: Instead of validating the size a lot of times, it would be better to have a function with an if for it, like the other ones. `getPaddingPadding` and `getFontSize`, for example.

nyan07:
> packages/components/src/index.ts

No. because we have just one package for all the components. If you look in this file `packages/components/src/index.ts, you will see that all components are also there. If you don't do that, you will create a private component, which can be used only on this repo.

packages/components/src/components/Badge/Badge.tsx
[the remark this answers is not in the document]
nyan07: Something like: "A visual label representing a status, property, or some other metadata."

packages/components/src/components/Badge/Badge.tsx
richbachman: Can a badge be disabled? It's just a div, right? It shouldn't be clickable, so no need to be disabled.

richbachman (changes requested):

packages/components/src/components/Badge/Badge.tsx
apvale: They have the type disabled here https://www.figma.com/file/0KtV7fzjug9hL3XOLzqn1l/%5B-WIP-%5D-%F0%9F%AA%90-Pluto-Components?node-id=255%3A617 
Why can't a badge be disabled? If it's not clickable, it is disabled, right?

blumaa: Also I asked in the design channel and they said there is a disabled state: https://localyzeteam.slack.com/archives/C03DUMP3XD4/p1666624596425449

packages/components/src/components/Badge/Badge.tsx
blumaa: It seems it's always on the left side. https://localyzeteam.slack.com/archives/C03DUMP3XD4/p1666624596425449

packages/components/src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+} as ComponentMeta<typeof Badge>;
+
+const Template: ComponentStory<typeof Badge> = (args) => <Badge {...args} />;
+
+export const Default = Template.bind({});
+Default.args = {
+  children: "Primary badge",
+};
+
+export const Disabled = Template.bind({});
```
nyan07, line 17: We are still calling it disabled here.

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,145 @@
...
+  | "primary"
+  | "success"
+  | "warning";
+
+export interface BadgeProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, "color"> {
+  /** The contents of the badge. Can be text. */
+  children: NonNullable<React.ReactNode>;
+  /** Sets the badge to be disabled. This is still being discussed with design whether we need the disabled prop or not. */
+  disabled?: boolean;
```
nyan07, line 22: I think @richbachman left a comment about the disabled prop, which is not solved yet.

nyan07: You probably need to talk with the design team again because I think it is on the Figma layout. It should be another variant, even though I'm not sure if `disabled` is a good name for it. We already have inactive, which is weird. Why do we need two different colors for the same thing? :thinking:

packages/components/src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,116 @@
...
+export interface BadgeProps
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, "color"> {
+  /** The contents of the badge. Can be text. */
+  children: NonNullable<React.ReactNode>;
+  /** Icon to be added on the left of the content. */
+  icon?: IconNames;
+  /** Sets the size of the badge. */
+  size?: BadgeSizeOptions;
+  /** Sets the variant of the badge. */
+  variant?: BadgeVariantOptions;
```
nyan07, line 21: I think that would be better to call it `color` as we talked on slack.

richbachman: Yeah. I agree. We can save `variant` if we ever add larger badge sizes.

richbachman:
@blumaa I missed that we need to the changeset files to this PR before merging. Ping me if you'd like a quick walkthrough.

packages/theme/src/theme/default.ts
```diff
@@ -122,6 +122,7 @@ export const theme = {
     space0: "0rem", // 0px
     space10: "0.125rem", // 2px
     space20: "0.25rem", // 4px
+    space25: "0.4rem", // 6px
```
nyan07, line 125: Not sure if we should add this token. Do we need to use it only in this component or across the app? Because I think the design team hasn't added it as a token. What do you think @richbachman?

nyan07 (approved):

PR #147, opened 2022-10-26
feat(components): add list

nyan07:
## Description of the change

This PR adds a list of items. It could be ordered or unordered. 

## Testing the change

- Run storybook

## Type of change

### Development

### Code review

image


packages/components/src/components/List/UnorderedList.tsx
```diff
@@ -0,0 +1,49 @@
+import styled from "@xstyled/styled-components";
+import React, { ReactElement } from "react";
+import { Box } from "../../primitives/Box";
+import { ListItemProps } from "./ListItem";
+
+export type ListFontColors = "colorText" | "colorTextLink";
+export interface UnorderedListProps
+  extends Omit<React.HTMLAttributes<HTMLUListElement>, "color"> {
+  /** Changes the color of the list items. */
+  color?: ListFontColors;
```
apvale, line 10: Should this type be part of the ListItem? 🤔

packages/components/src/components/List/OrderedList.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { ReactElement } from "react";
+import { Box } from "../../primitives/Box";
+import { ListItemProps } from "./ListItem";
+
+export interface OrderedListProps
+  extends Omit<React.HTMLAttributes<HTMLOListElement>, "color"> {
+  /** The list items */
+  children: ReactElement<ListItemProps> | ReactElement<ListItemProps>[];
```
apvale, line 8: Why doesn't this component have the colors with that two tokens as type prop?

packages/components/src/components/List/UnorderedList.tsx
```diff
@@ -0,0 +1,49 @@
...
+/** A list of items with bullet points */
+const InnerUnorderedList = styled(Box.ul)`
+  margin-left: space30;
+
+  li {
+    position: relative;
+    padding-left: space30;
+
+    &::before {
+      content: "•";
```
apvale, line 25: Why using just `ul` is not enough to add the bullet points?

packages/components/src/components/List/UnorderedList.tsx
richbachman: Why does a list need a color?

packages/components/src/components/List/UnorderedList.tsx
nyan07: The bullet points have their spacing and size defined by the font size. In our case, it is really small, so we can't control it. It is possible to control the size using the pseudo element `::marker`, but not the spacing.

packages/components/src/components/List/UnorderedList.tsx
nyan07: When we have an anchor list the bullet points should have a different color. But I just realized that it can follow the container color. So, I updated the stories.

packages/components/src/components/List/OrderedList.tsx
nyan07: I just removed the color https://github.com/Localitos/pluto/pull/147#discussion_r1005886838

packages/components/src/components/Callout/Callout.stories.tsx
```diff
@@ -18,9 +20,11 @@ Default.args = {
   children: (
     <Box.div>
       <Text.p>This is a default message.</Text.p>
-      <Text.a color="colorTextInfo" href="#">
-        Text Link
-      </Text.a>
+      <UnorderedList color="colorTextLink">
```
richbachman, line 23: Should this color prop be removed?

packages/components/src/components/List/UnorderedList.tsx
```diff
@@ -0,0 +1,42 @@
+import styled from "@xstyled/styled-components";
+import React, { ReactElement } from "react";
+import { Box } from "../../primitives/Box";
+import { ListItemProps } from "./ListItem";
+
+export type ListFontColors = "colorText" | "colorTextLink";
```
apvale, line 6: This can be removed, right?

packages/components/src/components/List/UnorderedList.tsx
```diff
@@ -0,0 +1,42 @@
...
+export interface UnorderedListProps
+  extends Omit<React.HTMLAttributes<HTMLUListElement>, "color"> {
```
apvale, line 8: Also, we don't need to do this anymore.

packages/components/src/components/List/OrderedList.tsx
```diff
@@ -0,0 +1,41 @@
+import styled from "@xstyled/styled-components";
+import React, { ReactElement } from "react";
+import { Box } from "../../primitives/Box";
+import { ListItemProps } from "./ListItem";
+
+export interface OrderedListProps
+  extends Omit<React.HTMLAttributes<HTMLOListElement>, "color"> {
```
apvale, line 7: I think we can remove this.

packages/components/src/components/List/ListItem.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { ReactNode } from "react";
+import { Box } from "../../primitives/Box";
+
+export interface ListItemProps
+  extends Omit<React.HTMLAttributes<HTMLLIElement>, "color"> {
```
apvale, line 5: And also remove it from here.

apvale (approved):

packages/components/src/components/List/UnorderedList.tsx
nyan07: We need it because the `Box.ul` has color and also the `HTMLUListElement` and they are incompatible.

packages/components/src/components/List/OrderedList.tsx
nyan07: Same here https://github.com/Localitos/pluto/pull/147#discussion_r1006881417

packages/components/src/components/List/ListItem.tsx
nyan07: Same here https://github.com/Localitos/pluto/pull/147#discussion_r1006881417

apvale (approved):

PR #158, opened 2022-10-28
feat(design-tokens): add design tokens package

richbachman:
## Description of the change

This sets up the initial design tokens package. Right now we're exporting the tokens as css, scss, and js. The js isn't quite in the format we need yet, but we can tackle that in a future PR. The css will be useful for styling the reporting ag-grid instance in the client app.

Eventually we should be able to use the js file in our theme package, and therefore have one source of truth for colors, spacing, etc.

I'm also having trouble with how to test this right now. I spent a few hours trying snapshots, with no luck. The error I'm running into is:

```
Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/richbachman/Code/Localitos/pluto/packages/design-tokens/dist/js/variables.js:6
    export const borderRadius10 = "4px";
    ^^^^^^

    SyntaxError: Unexpected token 'export'

    > 1 | import styles from "../../dist/js/variables";
        | ^
      2 |
      3 | describe("Design Tokens", () => {
      4 |   it("should generate js variables", () => {

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1796:14)
      at Object.<anonymous> (packages/design-tokens/src/__tests__/index.test.tsx:1:1)
```

## Testing the change

## Type of change

### Development

### Code review


apvale (approved):

apvale:
Reminder: remove the test that is not working before merging 🙏

PR #184, opened 2022-11-02
feat(components): add ghost button variant

nyan07:
## Description of the change
- Renames the `text` variant to `ghost`
- Enables `iconOnly` version
- Refactores stories to have all variants statuses 
- Fixes active and focus states

image

## Testing the change
- Run storybook

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Button/Button.stories.tsx
+++ b/packages/components/src/components/Button/Button.stories.tsx
@@ -1,28 +1,42 @@
 import type { ComponentMeta, ComponentStory } from "@storybook/react";
+import map from "lodash/map";
 import React from "react";
-import { Button } from "./Button";
+import { Box } from "../../primitives/Box";
+import { Button, ButtonProps } from "./Button";
 
 export default {
   component: Button,
   title: "Components/Button",
 } as ComponentMeta<typeof Button>;
 
-const Template: ComponentStory<typeof Button> = (args) => <Button {...args} />;
+const variants = ["primary", "secondary", "ghost"];
+
+const Template: ComponentStory<typeof Button> = (args) => (
+  <Box.div alignItems="center" display="flex" gap="space30">
+    {map(variants, (variant: ButtonProps["variant"]) => {
+      return (
+        <Box.div padding="space60">
+          <Button {...args} variant={variant} />
+        </Box.div>
+      );
+    })}
+  </Box.div>
+);
 
 export const Default = Template.bind({});
 Default.args = {
-  children: "Primary button",
+  children: "Default button",
 };
 
 export const Large = Template.bind({});
 Large.args = {
-  children: "Large primary button",
+  children: "Large button",
   size: "large",
 };
 
 export const Disabled = Template.bind({});
 Disabled.args = {
-  children: "Primary disabled button",
+  children: "Disabled button",
   disabled: true,
 };
 
@@ -32,20 +46,18 @@ Loading.args = {
   loading: true,
 };
 
-export const Secondary = Template.bind({});
-Secondary.args = {
-  leadingIcon: "SunIcon",
-  size: "large",
-  children: "Secondary button",
-  variant: "secondary",
+export const WithLeadingIcon = Template.bind({});
+WithLeadingIcon.args = {
+  children: "Button with icon",
+  size: "small",
+  leadingIcon: "ArrowTopRightOnSquareIcon",
 };
 
-export const Text = Template.bind({});
-Text.args = {
-  leadingIcon: "ArrowLongLeftIcon",
+export const WithTrailingIcon = Template.bind({});
+WithTrailingIcon.args = {
+  children: "Button with icon",
   size: "large",
-  children: "Text",
-  variant: "text",
+  trailingIcon: "ArrowTopRightOnSquareIcon",
 };
 
 export const AsLink = Template.bind({});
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Button/Button.stories.tsx
```diff
@@ -1,28 +1,42 @@
...
+import { Box } from "../../primitives/Box";
+import { Button, ButtonProps } from "./Button";
 
 export default {
   component: Button,
   title: "Components/Button",
 } as ComponentMeta<typeof Button>;
 
-const Template: ComponentStory<typeof Button> = (args) => <Button {...args} />;
+const variants = ["primary", "secondary", "ghost"];
```
apvale, line 12: Why not import the Button variants here?

packages/components/src/components/Button/Button.tsx
```diff
@@ -5,16 +5,16 @@ import { Box } from "../../primitives/Box";
...
 
 type IconNames = keyof typeof HeroOutlineIcons;
 
 export interface ButtonProps
   extends Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, "color"> {
   /** Sets the render element of the component. Either 'a' or 'button'.*/
-  as?: "a" | "button";
-  /** Add a description comment for each prop. */
-  children: NonNullable<React.ReactNode>;
+  as?: "a" | "button" | "label";
```
apvale, line 15: Is this semantically correct? Having a button as a label?

packages/components/src/components/Button/Button.tsx
```diff
@@ -33,22 +33,34 @@ export interface ButtonProps
   variant: ButtonVariantOptions;
```
apvale, line 33: Can you move the mandatory props to the top, please?

packages/components/src/components/Button/Button.tsx
```diff
@@ -33,22 +33,34 @@ export interface ButtonProps
...
   /** True to have a full width button, false otherwise. */
   fullWidth?: boolean;
+
```
apvale, line 36: Maybe adding a desc for this prop?

packages/components/src/components/Button/Button.tsx
```diff
@@ -109,6 +115,31 @@ const getIcon = (iconName: IconNames, size: ButtonSizeOptions) => {
...
+      paddingTop: size === "large" ? "space40" : "space30",
+      paddingLeft: size === "large" ? "space40" : "space30",
+      paddingRight: size === "large" ? "space40" : "space30",
+    };
+  }
+  return {
+    paddingBottom: size === "large" ? "space40" : "space30",
+    paddingTop: size === "large" ? "space40" : "space30",
+    paddingLeft: size === "large" ? "space50" : "space40",
+    paddingRight: size === "large" ? "space50" : "space40",
```
apvale, line 139: I was checking on Figma, and if the **small button** has an Icon (leading or trailing), it has only the left or right padding different from the others:

image

https://www.figma.com/file/0KtV7fzjug9hL3XOLzqn1l/%5B-WIP-%5D-%F0%9F%AA%90-Pluto-Components?node-id=149%3A2308

packages/components/src/components/Button/Button.stories.tsx
nyan07: Because it is a type and can't be used as a value.

packages/components/src/components/Button/Button.tsx
richbachman: I don't think a button should ever render as a label.

nyan07: The only reason to add it is to use it on the upload component. To be accessible, the input should have a label, but in our case, the label looks like a button. I don't think it will be used this way in another place. Do we have another way to style a label likes a button?

richbachman:
I really think we should be using `children` here to have the icon. That way it's accessible. Right now, the leading icon is decorative, and has no title. Since `aria-label` isn't mandatory it will be easily forgotten, and we'll have a bunch of inaccessible buttons.

nyan07:
> I really think we should be using `children` here to have the icon. That way it's accessible. Right now, the leading icon is decorative, and has no title. Since `aria-label` isn't mandatory it will be easily forgotten, and we'll have a bunch of inaccessible buttons.

@richbachman I understand, but how could I change the icon size according to the button size? If we leave it up to the developer it is easy to use incorrectly.

packages/components/src/components/Button/Button.tsx
apvale: I never saw it, but I just did a google and found that MUI is using buttons as labels as well https://mui.com/material-ui/react-button/#upload-button. So, I think it's a usual thing to do.

richbachman:
> @richbachman I understand, but how could I change the icon size according to the button size? If we leave it up to the developer it is easy to use incorrectly.

Changing the icon size is the lesser of the two evils here. I'd rather have a wrongly sized icon over an inaccessible button. Wrongly sized icons are much easier to catch visually and correct.

nyan07:
> @nyan07 Changing the icon size is the lesser of the two evils here. I'd rather have a wrongly sized icon over an inaccessible button. Wrongly sized icons are much easier to catch visually and correct.

@richbachman, the size is just an example. But there are other styles we need to deal with according to the button state. And if we throw an exception as you did on the icon component?

richbachman (approved):
LGTM!

PR #193, opened 2022-11-07
feat(components): added Modal component

richbachman:
## Description of the change

This PR adds the initial release of the modal component. The component uses the [AriaKit](https://ariakit.org/components/dialog) primitive to handle the modal functionality. Currently I'm leaving the modal API pretty open, as in straight exports from AriaKit. This is mainly to get a feel for how this modal will work with the current application. We will be adding a more dev friendly modal in the future where the api will be more refined. I have also not included any animations in this first version. Because of the way the primitive is built, we'll need to setup an animation library like Framer first in Pluto. I'd like to spend more time researching Framer or another option first before adding animations in this PR.

This also includes an update to the theme to include the backdrop styles.

Screen Shot 2022-11-07 at 13 32 06

## Testing the change

- [ ] Check out the storybook stories

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Modal/Modal.tsx
+++ b/packages/components/src/components/Modal/Modal.tsx
@@ -1,0 +1,40 @@
+import React from "react";
+import { Dialog } from "ariakit/dialog";
+import type { DialogProps } from "ariakit";
+import { Box } from "../../primitives/Box";
+
+export interface ModalProps extends DialogProps {
+  /** The contents of the modal. */
+  children: NonNullable<React.ReactNode>;
+}
+
+/** A Modal is a page overlay that displays information and blocks interaction with the page until an action is taken or the Modal is dismissed. */
+const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
+  ({ children, ...props }, ref) => {
+    return (
+      <Box.div
+        as={Dialog}
+        backgroundColor="colorBackground"
+        borderRadius="borderRadius30"
+        display="flex"
+        flexDirection="column"
+        left="50%"
+        maxHeight="90vh"
+        maxWidth="47.75rem"
+        position="fixed"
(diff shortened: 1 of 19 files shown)


packages/components/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { Dialog } from "ariakit/dialog";
+import type { DialogProps } from "ariakit";
+import { Box } from "../../primitives/Box";
+
+export interface ModalProps extends DialogProps {
+  children: NonNullable<React.ReactNode>;
+}
+
+const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
```
RohitRajendran, line 10: Missing a jsdoc description for Modal

packages/components/src/components/Modal/ModalHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+      >
+        <Box.div>{children}</Box.div>
+        <Box.div>
+          <Button
+            aria-label="Close modal"
+            as={DialogDismiss as unknown as keyof JSX.IntrinsicElements}
+            iconOnly
+            leadingIcon="XMarkIcon"
+            variant="ghost"
+          />
```
RohitRajendran, line 33: Instead of doing it this way and having to mess with the cast of DialogDismiss, you can actually pass `Button` to the `as` prop for `DialogDismiss`:

```       
<DialogDismiss
    as={Button}
    aria-label="Close modal"
    iconOnly
    leadingIcon="XMarkIcon"
    variant="ghost"
  />

richbachman: Good catch!

PR #205, opened 2022-11-10
Feat/dropzone

blumaa:
## Description of the change

This PR adds the dropzone for file uploading. I still need to figure out a way to load the variants in storybook without dropping a file. Suggestions? Also, I need to add some tests as well. Any ideas for those are welcome. 
Screenshot 2022-11-15 at 19 32 38

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
     "packages/**"
   ],
   "dependencies": {
-    "@tanstack/react-table": "^8.5.30"
+    "@tanstack/react-table": "^8.5.30",
+    "react-dropzone": "^14.2.3"
   },
   "devDependencies": {
     "@changesets/changelog-github": "0.4.7",
--- a/packages/components/src/components/Dropzone/Dropzone.stories.tsx
+++ b/packages/components/src/components/Dropzone/Dropzone.stories.tsx
@@ -1,57 +1,44 @@
 import type { ComponentMeta, ComponentStory } from "@storybook/react";
-import React, { useState } from "react";
-import Uppy from "@uppy/core";
+import React from "react";
+import { FileWithPath } from "react-dropzone";
 import { Dropzone } from "./Dropzone";
 
 const resolveAfter2Seconds = () => {
   return new Promise((resolve) => {
     setTimeout(() => {
-      resolve("resolved");
+      resolve("Document uploaded successfully");
     }, 2000);
   });
-}
-
-const sendDocument = async (
-  file: File,
-  onProgress?: (progressPercentage: number) => void
-) => {
-  console.log("calling");
-  const result = await resolveAfter2Seconds();
-  console.log(result);
-  return result;
 };
 
-// const sendDocument = async (
-//   file: File,
-//   onProgress?: (progressPercentage: number) => void
-// ) => {
-//   // let count = 0;
+const sendDocument = async (
+  file: FileWithPath,
+  onProgress?: (progressPercentage: number) => void
+) => {
+  let current = 0;
+  let interval = null;
 
-//   // setTimeout(() => {
-//   //   const intervalID = setInterval(() => {
-//   //     if (count === 100) {
-//   //       clearInterval(intervalID);
-//   //     } else {
-//   //       count++;
-//   //     }
-//   //   }, 100);
-//   // }, 2000);
+  const useCallback = () => {
+    if (onProgress !== undefined) {
+      onProgress(current);
+    }
+    const nextCount = current++;
+    if (nextCount === 100) {
+      clearInterval(interval);
+    }
+  };
 
-//   // if (onProgress !== undefined) {
-//   //   onProgress(count);
-//   // }
+  interval = setInterval(useCallback, 10);
 
-//   return new Promise(function (resolve, reject) {
-//     setTimeout(() => {
-//       return reject(new Error("It broke"));
-//     }, 5000);
-//   });
-// };
+  if (current < 100) {
+    return await resolveAfter2Seconds();
+  }
+};
 
-const cancelDocumentUpload = () => {
+const cancelDocumentUpload = (name: string) => {
   return new Promise(function (resolve, reject) {
     setTimeout(() => {
-      return reject(new Error("It broke"));
+      return reject(new Error(`${name} failed to upload`));
     }, 1000);
   });
 };
@@ -62,13 +49,16 @@ export default {
 } as ComponentMeta<typeof Dropzone>;
 
 const Template: ComponentStory<typeof Dropzone> = (args) => (
-  <Dropzone {...args} />
+  <Dropzone
+    {...args}
+    cancelDocumentUpload={cancelDocumentUpload}
+    sendDocument={sendDocument}
+  />
 );
 
 export const Default = Template.bind({});
-Default.args = {};
 
-export const WithDocumentUpload = (): React.ReactElement => (
+export const WithDocumentUploadSuccess = (): React.ReactElement => (
   <Dropzone
     cancelDocumentUpload={cancelDocumentUpload}
     sendDocument={sendDocument}
--- a/packages/components/src/components/Dropzone/Dropzone.stories.tsx
+++ b/packages/components/src/components/Dropzone/Dropzone.stories.tsx
@@ -1,0 +1,37 @@
+import type { ComponentMeta, ComponentStory } from "@storybook/react";
+import React from "react";
+import { Dropzone } from "./Dropzone";
+
+export default {
+  component: Dropzone,
+  title: "Components/Dropzone",
+} as ComponentMeta<typeof Dropzone>;
+
+const Template: ComponentStory<typeof Dropzone> = (args) => (
+  <Dropzone {...args} />
+);
+{
+  /* <Dropzone fileType="foo" onSuccess={() => console.log('yay')} /> */
+}
+
+export const Simple = Template.bind({});
+Simple.args = {};
+
+// User
+
+/* Potential props */
+/* FileTypes */
+/* OnSuccess */
+// This is the function that will send the file to an endpoint
+
+// Notes on Dropzone
+
+// Question about the icons? How do I Type them?
+
+// How do we get a loading state?
+
+// How do I combine the getBackgroundColors and getBorderColors functions?
+
+// How do I get the progress bar displaying?
+
+// What exactly is the focus and what do I need from the input?
--- a/packages/components/src/components/Dropzone/Dropzone.stories.tsx
+++ b/packages/components/src/components/Dropzone/Dropzone.stories.tsx
@@ -1,6 +1,60 @@
 import type { ComponentMeta, ComponentStory } from "@storybook/react";
-import React from "react";
+import React, { useState } from "react";
+import Uppy from "@uppy/core";
 import { Dropzone } from "./Dropzone";
+
+const resolveAfter2Seconds = () => {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve("resolved");
+    }, 2000);
+  });
+}
+
+const sendDocument = async (
+  file: File,
+  onProgress?: (progressPercentage: number) => void
+) => {
+  console.log("calling");
+  const result = await resolveAfter2Seconds();
+  console.log(result);
+  return result;
+};
+
+// const sendDocument = async (
+//   file: File,
+//   onProgress?: (progressPercentage: number) => void
+// ) => {
+//   // let count = 0;
+
+//   // setTimeout(() => {
+//   //   const intervalID = setInterval(() => {
+//   //     if (count === 100) {
+//   //       clearInterval(intervalID);
+//   //     } else {
+//   //       count++;
+//   //     }
+//   //   }, 100);
+//   // }, 2000);
+
+//   // if (onProgress !== undefined) {
+//   //   onProgress(count);
+//   // }
+
+//   return new Promise(function (resolve, reject) {
+//     setTimeout(() => {
+//       return reject(new Error("It broke"));
+//     }, 5000);
+//   });
+// };
+
+const cancelDocumentUpload = () => {
+  return new Promise(function (resolve, reject) {
+    setTimeout(() => {
+      return reject(new Error("It broke"));
+    }, 1000);
+  });
+};
 
 export default {
   component: Dropzone,
@@ -10,28 +64,13 @@ export default {
 const Template: ComponentStory<typeof Dropzone> = (args) => (
   <Dropzone {...args} />
 );
-{
-  /* <Dropzone fileType="foo" onSuccess={() => console.log('yay')} /> */
-}
 
-export const Simple = Template.bind({});
-Simple.args = {};
+export const Default = Template.bind({});
+Default.args = {};
 
-// User
-
-/* Potential props */
-/* FileTypes */
-/* OnSuccess */
-// This is the function that will send the file to an endpoint
-
-// Notes on Dropzone
-
-// Question about the icons? How do I Type them?
-
-// How do we get a loading state?
-
-// How do I combine the getBackgroundColors and getBorderColors functions?
-
-// How do I get the progress bar displaying?
-
-// What exactly is the focus and what do I need from the input?
+export const WithDocumentUpload = (): React.ReactElement => (
+  <Dropzone
+    cancelDocumentUpload={cancelDocumentUpload}
+    sendDocument={sendDocument}
+  />
+);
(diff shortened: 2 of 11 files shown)


packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+import type { SystemProp, Theme } from "@xstyled/styled-components";
+import { FileWithPath, useDropzone } from "react-dropzone";
+import forEach from "lodash/forEach";
+import { Text } from "../../primitives/Text";
+import { ProgressBar } from "../ProgressBar";
+import { Box } from "../../primitives/Box";
+import { Icon, IconProps } from "../Icon";
+import { CancelUploadButton } from "./CancelUploadButton";
+
+const MAX_FILE_SIZE = 50 * 1024 * 1024;
```
richbachman, line 11: Why the multiplication here?

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+type DropzoneProps = {
+  /* These props are being saved for V2 */
+  /** The maximum amount allowed for file size. In the format of number?. */
+  // MaxFileSize: number;
+  /** Are multiple file uploads allowed? Boolean */
+  // MultipleFiles: boolean;
```
richbachman, line 18: Lets go ahead and remove these. We can add them back in when ready.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+            isDragReject,
+          })}
+          ref={ref}
+        >
+          <input {...getInputProps()} />
+          {getIcon(status, isDragActive, isDragAccept, isDragReject)}
+
+          {!isDragActive && status === "loading" && (
+            <>
+              {acceptedFiles[0].name}
```
richbachman, line 185: What happens if there are multiple file types that need to be uploaded. This look like it checks just the first type.

blumaa: This is just the first file in the array. If there are more files we could map over them and call sendDocument on them?

packages/components/src/components/Dropzone/Dropzone.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { render } from "@testing-library/react";
+import React from "react";
+import { Dropzone } from "./Dropzone";
+
+describe("<Dropzone />", () => {
+  it("renders correctly", () => {
+    render(<Dropzone />);
```
nyan07, line 7: The `expect` is missing.

package.json
```diff
@@ -28,7 +28,10 @@
     "packages/**"
   ],
   "dependencies": {
-    "@tanstack/react-table": "^8.5.15"
+    "@tanstack/react-table": "^8.5.15",
+    "@uppy/aws-s3": "^3.0.4",
```
nyan07, line 32: Shouldn't uppy be only on dev dependencies?

jest.config.js
```diff
@@ -25,7 +25,9 @@ module.exports = {
       "<rootDir>/.jest/file-mock.ts",
   },
   setupFilesAfterEnv: ["<rootDir>/.jest/jest-setup.ts"],
-  transformIgnorePatterns: ["node_modules/"],
+  transformIgnorePatterns: [
```
nyan07, line 28: Could you test if you still need this with uppy on the dev dependencies, please?

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+          .catch(() => {
+            setStatus("uploadError");
+          });
+      },
+      onDropRejected: (fileRejections) => {
+        setStatus("error");
+        forEach(fileRejections, (file) => {
+          forEach(file.errors, (err) => {
+            if (err.code === "file-too-large") {
+              setErrors("File must be less than 50mb.");
```
nyan07, line 145: As pluto is opensource, what do you think about having a default max file size and allowing engineers to set the max they want as an optional prop? So, if someone else uses it, the component is it'll reusable. And not fixed for our needs.

packages/components/src/components/Dropzone/Dropzone.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+const resolveAfter2Seconds = () => {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve("Document uploaded successfully");
+    }, 2000);
+  });
+};
+
+const sendDocument = async (
+  file: FileWithPath,
```
nyan07, line 15: What do you think about exposing our own type? I think this way people will always import something from lib we are using. Instead the could import something like: 

```js
import { Dropzone, FileWithPath } from "./Dropzone";
```

packages/components/src/components/Dropzone/Dropzone.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onProgress?: (progressPercentage: number) => void
+) => {
+  let current = 0;
+  let interval = null;
+
+  const useCallback = () => {
+    if (onProgress !== undefined) {
```
nyan07, line 22:
```suggestion
    if (!!onProgress) {
```

packages/components/src/components/Dropzone/Dropzone.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export default {
+  component: Dropzone,
+  title: "Components/Dropzone",
+} as ComponentMeta<typeof Dropzone>;
+
+const Template: ComponentStory<typeof Dropzone> = (args) => (
+  <Dropzone
+    {...args}
+    cancelDocumentUpload={cancelDocumentUpload}
```
nyan07, line 54: Ir our case, we will use dropzone to upload documents, but don't forget that pluto is an open-source project, and people could use it to upload pictures, for example. So, I would suggest creating a prop `cancelUpload` which receives a boolean, and another prop, `onCancelUpload` to have it as a callback.

packages/components/src/components/Dropzone/Dropzone.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    sendDocument={sendDocument}
```
nyan07, line 55: I think we need to rethink also the `sendDocument` prop.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+    onProgress?: (progressPercentage: number) => void
+  ) => Promise<unknown>;
+  /** This is the function that gets triggered when a file is dropped */
+  cancelDocumentUpload: (name: string) => Promise<unknown>;
+  /** The file types allowed. In the format of an object. */
+  fileTypes?: Record<string, string[]>;
+};
+/**/
+
+const getVariants = (
```
nyan07, line 31: I don't think we have variants for dropzone. Maybe would be better to call it `getColors`.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+                maxWidth="300px"
+                w="100%"
+              >
+                <Box.div maxWidth="280px" w="100%">
+                  <ProgressBar value={progress} />
+                </Box.div>
+                <CancelUploadButton
+                  onClick={(event) => {
+                    event.stopPropagation();
+                    cancelDocumentUpload(acceptedFiles[0].name).then(() =>
```
nyan07, line 199: Could we upload more than one file? What happens here if we try to add more files?

packages/components/src/components/Dropzone/Dropzone.tsx
sliminas: Probably taken from my POC code. Is there an easier and more readable way to get 50 MB in JS?
I guess there is nothing like in Rails where you can have `50.megabytes`.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+      borderColor: "colorBorderSuccess",
+    };
+  }
+  if (isDragAccept) {
+    return {
+      backgroundColor: "colorBackgroundInfo",
+      borderColor: "colorBorderPrimary",
+    };
+  }
+  if (isDragReject || status === "error" || status === "uploadError") {
```
sliminas, line 58: Maybe a bit shorter?
```suggestion
  if (isDragReject || ["error", "uploadError"].includes(status)) {
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,252 @@
...
+  let iconColor: IconProps["color"] = "colorIconInfo";
+  if ((!isDragActive && !isDragReject) || status === "success") {
+    icon = "CheckCircleIcon";
+    iconColor = "colorIconSuccess";
+  }
+  if (isDragAccept || status === "default" || status === "loading") {
+    icon = "CloudArrowUpIcon";
+    iconColor = "colorIconInfo";
+  }
+  if (isDragReject || status === "error" || status === "uploadError") {
```
sliminas, line 92: We do we need to check both the `*drag*` variables and the `status`?
Aren't the drag variables there exactly for representing the state?

packages/components/src/components/Dropzone/Dropzone.stories.tsx
blumaa: Can you say more here @nyan07 ? How so?

packages/components/src/components/Dropzone/Dropzone.tsx
blumaa: V1 only allows the uploading of one file. V2 we might accept multiple files. In that case we'll have to map over the acceptedFiles array and sendDocument on each file. Make sense?

packages/components/src/components/Dropzone/Dropzone.tsx
blumaa: The problem is that isDragReject only checks the file type when there is a drag over. So, when we want to check if the file size is too big or whether there was a file upload error, we need the 'error' and 'uploadError' states.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+import { CancelUploadButton } from "./CancelUploadButton";
+
+const MAX_FILE_SIZE = 52_428_800;
+
+type DropzoneProps = {
+  error: string | undefined;
+  /** This is the function that gets triggered when a file is dropped */
+  onDrop: (files: FileWithPath[]) => void;
+  /** This is the function that gets triggered when the cancel button is clicked */
+  onCancelClick: (files: FileWithPath[]) => void;
```
richbachman, line 20: Thoughts on just changing this to onCancel?

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+  /** The file types allowed. In the format of an object. */
+  fileTypes?: Record<string, string[]>;
+  /** What is the max file size allowed? Default is 50mb. Number of bytes */
```
richbachman, line 23:
```suggestion
  /** The max file size allowed. Default is 50mb. Number of bytes */
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+  maxFileSize?: number;
+  progress?: number;
```
richbachman, line 25: Just need a comment for this prop.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const [dropZoneError, setDropZoneError] = useState(error);
+    const defaultState = !dropZoneError && !progress;
+    const errorState = dropZoneError;
+    const isLoading = !!progress && progress > 0 && progress < 100;
+    const successState = progress === 100 && !dropZoneError;
+
+    // console.log("defaultState", defaultState);
+    // console.log("errorState", errorState);
+    // console.log("isLoading", isLoading);
+    // console.log("successState", successState);
```
richbachman, line 112: Reminder to remove these commented lines.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+        onDrop(acceptedFiles);
+      },
+      onDropRejected: (fileRejections) => {
+        forEach(fileRejections, (file) => {
+          forEach(file.errors, (err) => {
+            if (err.code === "file-too-large") {
+              setDropZoneError("File must be less than 50mb.");
+            }
+            if (err.code === "file-invalid-type") {
+              setDropZoneError("Wrong file type. PDF format only.");
```
richbachman, line 131: Since we're allowing multiple file types, we may want to reword this error. We are allowing multiple file types right?

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+            isDragAccept,
+            isDragReject
+          )}
+          alignItems="center"
+          borderRadius={"borderRadius30"}
+          borderStyle={"borderDashed"}
+          borderWidth={"borderWidth10"}
+          display="flex"
+          flexDirection="column"
+          padding={"space70"}
```
richbachman, line 155:
```suggestion
          padding="space70"
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+                  onClick={(event) => {
+                    event.stopPropagation();
+                    onCancelClick(acceptedFiles);
+                  }}
+                />
+              </Box.div>
+            </>
+          )}
+          {successState && (
+            <Text.span color="colorTextStronger" fontSize={"fontSize20"}>
```
richbachman, line 186:
```suggestion
            <Text.span color="colorTextStronger" fontSize="fontSize20">
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+              {acceptedFiles.join(", ")}
+            </Text.span>
+          )}
+          {errorState && (
+            <Text.span
+              color="colorTextStronger"
+              fontSize={"fontSize30"}
```
richbachman, line 194:
```suggestion
              fontSize="fontSize30"
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+              fontWeight="fontWeightRegular"
+            >
+              {dropZoneError}
+            </Text.span>
+          )}
+          {defaultState && (
+            <>
+              <Text.span
+                color="colorTextStronger"
+                fontSize={"fontSize30"}
```
richbachman, line 205:
```suggestion
                fontSize="fontSize30"
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+                fontWeight="fontWeightRegular"
+              >
+                Drag and drop or browse
+              </Text.span>
+              <Text.span color="colorText" fontSize={"fontSize20"}>
```
richbachman, line 210:
```suggestion
              <Text.span color="colorText" fontSize="fontSize20">
```

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,234 @@
...
+                File must be PDF format and no larger than 50MB
```
richbachman, line 211: Maybe I'm confused on v1 allowing multiple types. Please correct me if wrong.

packages/components/src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,253 @@
...
+          borderRadius="borderRadius30"
+          borderStyle="borderDashed"
+          borderWidth="borderWidth10"
+          display="flex"
+          flexDirection="column"
+          padding="space70"
+          {...getRootProps()}
+          ref={ref}
+        >
+          {/* eslint-disable-next-line jsx-a11y/no-redundant-roles */}
```
richbachman, line 178: Nit: Would you mind adding a comment to this one so we know why we're disabling it.

richbachman (approved):
Just that one small thing from me.

PR #221, opened 2022-11-14
feat(components): add Tooltip component

richbachman:
## Description of the change

This PR adds the initial release of the tooltip component. The component uses the [AriaKit](https://ariakit.org/components/tooltip) primitive to handle the functionality. Currently I'm leaving the tooltip API pretty open, as in straight exports from AriaKit. This is mainly to get a feel for how this tooltip will work with the current application. We will be adding a more dev friendly tooltip in the future where the api will be more refined. I have also not included any animations in this first version. Because of the way the primitive is built, we'll need to setup an animation library like Framer first in Pluto. I'd like to spend more time researching Framer or another option first before adding animations in this PR.

Screenshot 2022-11-14 at 16 15 48

## Testing the change

- [ ] Check out the storybook stories

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Tooltip/Tooltip.test.tsx
+++ b/packages/components/src/components/Tooltip/Tooltip.test.tsx
@@ -1,13 +1,13 @@
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import React from "react";
-import { Default as TooltipContainer } from "./Tooltip.stories";
+import { Default as TooltipComponent } from "./Tooltip.stories";
 
 const OPEN_TOOLTIP_TEXT = "Hover or focus on me";
 
 describe("<Tooltip />", () => {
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -0,0 +1,36 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+import { Default } from "./Tooltip.stories";
```
nyan07, line 4: Maybe using a descriptive name would be better to understand the tests :nail_care: 

```suggestion
import { Default as TooltipContainer } from "./Tooltip.stories";
```

nyan07 (approved):

PR #230, opened 2022-11-16
fix(components): hide file info on FileUploader if there is no file name

padulafacundo:
## Description of the change
When the status of the `FileUploader` component is `success` but there is no `fileName` we show `false` as file name and the file size/max file size. In this scenario we shouldn't show anything.

image

## Testing the change

This didn't come up in the stories because we don't pass the size (and because of how the `FileUploaderDescription` works we rendered an empty `span`, instead of not rendering anything at all). 
You can test this by adding `fileSize` or `maxFileSize` on the story and comparing `main` vs this branch.

## Type of change

### Development

### Code review


packages/components/src/components/FileUploader/FileUploader.test.tsx
```diff
@@ -100,7 +100,13 @@ describe("<FileUploader />", () => {
...
       expect(screen.getByRole("link", { name: "Visa" })).toBeInTheDocument();
       expect(
-        screen.queryByText("No file uploaded • 1MB")
+        screen.queryByText("No file uploaded", { exact: false })
+      ).not.toBeInTheDocument();
+      expect(
+        screen.queryByText("1MB", { exact: false })
+      ).not.toBeInTheDocument();
+      expect(
+        screen.queryByText("2MB", { exact: false })
```
apvale, line 109: Why add these checks?

padulafacundo: To make sure we don't show _any_ kind of file information. Before we only had the "1MB" which is the `fileSize`, but if you see the screenshot in the description it's also possible that we incorrectly show the `maxFileSize`. If we had had the "2MB" check then we might have caught this. I figured I'd do different expectations (not exact) to avoid having all possible combinations

packages/components/src/components/FileUploader/FileUploader.tsx
```diff
@@ -108,7 +108,7 @@ const FileUploader = React.forwardRef<HTMLDivElement, FileUploaderProps>(
                 {!isMobile && status === "loading" && (
                   <FileUploaderProgressBar progress={progress} />
                 )}
-                {!isMobileUploading && (
+                {!isMobileUploading && (status !== "success" || fileName) && (
```
jvaladas, line 111: This logic can be hard to understand so my suggestion would be to move it to a variable like 

```javascript
  const isUploadedWithoutName = status === 'success' && !fileName;
```
```suggestion
                {(!isMobileUploading && !isUploadedWithoutName) && (
```

padulafacundo: Good idea!

packages/components/src/components/FileUploader/FileUploader.test.tsx
apvale: I don't think checking all of these texts is a good practice, but I understand the issue. We can't use a data-testid here, so, that's fine for me. But let's try to remove the `{ exact: false }` because we don't need them.

apvale (approved):

PR #237, opened 2022-11-21
feat(components): add Drawer component

richbachman:
## Description of the change

This PR adds the initial release of the drawer component. The component uses the [AriaKit](https://ariakit.org/components/dialog) primitive to handle the modal functionality. Currently I'm leaving the drawer API pretty open, as in straight exports from AriaKit. This is mainly to get a feel for how this drawer will work with the current application. We will be adding a more dev friendly drawer in the future where the api will be more refined. I have also not included any animations in this first version. Because of the way the primitive is built, we'll need to setup an animation library like Framer first in Pluto. I'd like to spend more time researching Framer or another option first before adding animations in this PR. The drawer also does not use a backdrop like the modal per the design, although clicking outside of the drawer will still close the drawer.

Screenshot 2022-11-21 at 12 53 42

## Testing the change

- [ ] Load storybook

## Type of change

### Development

### Code review


blumaa:
Cool about the animations library starting! A question: I know there is a framer and there is framer-motion. And they are related. Are you talking about them both in this context?

richbachman:
> Cool about the animations library starting! A question: I know there is a framer and there is framer-motion. And they are related. Are you talking about them both in this context?

@blumaa I think Framer changed the name to Framer Motion. https://www.framer.com/motion/

blumaa:
> > Cool about the animations library starting! A question: I know there is a framer and there is framer-motion. And they are related. Are you talking about them both in this context?
> 
> @blumaa I think Framer changed the name to Framer Motion. https://www.framer.com/motion/

Wow that is cool. It's been a while since I used it. :)

richbachman:
@blumaa I lied. Framer is a whole site builder thing like webflow. Framer motion is their open source animation library.

packages/components/src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,39 @@
...
+    return (
+      <Box.div
+        as={Dialog}
+        backdrop={false}
+        backgroundColor="colorBackground"
+        bottom={0}
+        boxShadow="shadowStrong"
+        display="flex"
+        flexDirection="column"
+        maxWidth="34.375rem"
```
apvale, line 23: Why don't we have the 550px in the DS? 🤔

packages/components/src/components/Drawer/DrawerBody.tsx
```diff
@@ -0,0 +1,43 @@
...
+        background="
+        linear-gradient(
+          white 30%,
+          rgba(255, 255, 255, 0)
+        ) center top,
+        radial-gradient(
+          farthest-side at 50% 0,
+          rgba(0, 0, 0, 0.2),
+          rgba(0, 0, 0, 0)
+        ) center top"
```
apvale, line 30: Shouldn't this be a token of the DS? I mean, I can't see this on [Figma](https://www.figma.com/file/uDEjU77uIrQ5T1csn8IvIJ/Pluto-Tokens?node-id=1%3A18&t=uoBTEB0BP2UvHwMs-1), for example.

packages/components/src/components/Drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,50 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+import { Default } from "./Drawer.stories";
+
+const OPEN_DRAWER_TEXT = "Open drawer";
+
+describe("<Drawer />", () => {
+  it("should open and close a drawer using the close drawer button", async () => {
+    render(<Default />);
```
apvale, line 10: Why use the component from the Stories? Shouldn't we use the component that we created in the `Drawer.tsx`? 🤔

packages/components/src/components/Drawer/Drawer.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    await userEvent.click(renderedOpenButton);
+    expect(screen.getByRole("dialog")).toBeInTheDocument();
+    expect(screen.getByRole("heading", { level: 2 })).toBeInTheDocument();
+
+    await userEvent.keyboard("{Escape}");
+    expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
+  });
+
+  it("should open a drawer and close it by clicking an element in the background", async () => {
+    render(<Default />);
```
apvale, line 39: I understand why you did that now. But I wonder if we are testing ariakit twice by doing these tests because of the close button.

apvale (commented):
Two questions about this:
1. Why add a component that is not present in the DS on Figma? I mean, I couldn't find it in this list https://www.figma.com/file/0KtV7fzjug9hL3XOLzqn1l/%5B-WIP-%5D-%F0%9F%AA%90-Pluto-Components?node-id=34%3A205&t=L5Y0xz2u7jSoeXT4-1. Unless exists a different one that I'm not aware of it 😅 
2. Shouldn't we also give the availability of changing the Drawer position? Now it's only on the Right, but if we want to show on the left?

packages/components/src/components/Drawer/DrawerBody.tsx
richbachman: Possibly. Right now its a one off use case to give show a shadow on scrolled content. If we find that we're reusing it more, we can add it in the future.

packages/components/src/components/Drawer/Drawer.test.tsx
richbachman: Using the component from stories allows us to test without using duplicate code. Storybook allows stories to be used for testing now, so we don't have to build another component just for tests. Here's an [article](https://storybook.js.org/docs/react/writing-tests/importing-stories-in-tests) for more information.

apvale (approved):

PR #304, opened 2022-12-09
fix(components): replace progress bar height with size variant

jvaladas:
## Description of the change

By replacing the height of the progress bar with a small/large size variant we simplify the usage of the component and prevent issues at later stages since the developers no longer have to worry about using the right height.

Screenshot 2022-12-09 at 16 56 56

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/ProgressBar/ProgressBar.tsx
+++ b/packages/components/src/components/ProgressBar/ProgressBar.tsx
@@ -4,29 +4,29 @@ import { Root, Indicator, ProgressProps } from "@radix-ui/react-progress";
 import { SystemProp, Theme } from "@xstyled/styled-components";
 import get from "lodash/get";
 
+type ProgressBarSizeOptions = "large" | "small";
+
 export interface ProgressBarProps {
   /** The progress value. */
   value: number;
-  /** The height of the progress bar */
-  h?: SystemProp<keyof Theme["space"], Theme>;
   /** The color of the progress indicator */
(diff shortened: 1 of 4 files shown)


blumaa (approved):
:)

packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -37,9 +37,9 @@ const StyledProgress = styled.div`
       theme.colors.colorBackgroundWeak
     )};
   border-radius: ${(props: StyledProgressProps) =>
-    get(theme.radii, String(props.radius), theme.radii.borderRadius10)};
+    get(theme.radii, String(props.radius), "")};
```
richbachman, line 40: Is this just passing an empty string? If so do we need it?

jvaladas: This is so lodash get can infer that we're using strings here. Removing it would require me to add something like

```suggestion
    get(theme.radii, String(props.radius)) as string};
```

jvaladas: The empty string here represents the default case but since radius is required this is what we get when the value is not found in theme.radii. Using the empty string also allows the style to fail instead failling silently by letting the developer pass a non existent radius here and the program still rendering everything perfectly

richbachman (approved):

PR #314, opened 2022-12-12
feat(components): add Select component

richbachman:
## Description of the change

This PR adds the `Select` component, tests, and stories.

Screenshot 2022-12-12 at 15 40 08

Along with `Select`, this also fixes a few minor bugs:

- [Box]: Added `auto` to `marginLeft` and marginRight` props.
- [InputBox]: Adjusted the focus border to correctly match the design.
- [Modal]: Potentially fixed the `nonce` bug.
- [Tooltip]: Potentially fixed the `nonce` bug.

There are also a couple new states added to the theme object:

- `activeItem` in order to style elements with `data-active-item`.
- `selected` in order to style elements with `aria-selected="true"`.

## Testing the change

- [ ] Check out storybook

## Type of change

### Development

### Code review


packages/components/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+        id={selectID}
+        items={selectItems}
+        name="select"
+      />
+      <HelpText id={helpTextID}>Please choose one of the values.</HelpText>
+    </Box.form>
+  );
+};
+
+export const WithFormik = (): JSX.Element => {
```
BrianLincoln, line 98: 💪

packages/components/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+      <Box.div marginTop="space70">
+        <Button type="submit" variant="primary">
+          Submit
+        </Button>
+      </Box.div>
+    </Box.form>
+  );
+};
+
+export const WithReactHookForm = (): JSX.Element => {
```
BrianLincoln, line 129: 💪

packages/components/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+        <Box.button
+          alignItems="center"
+          appearance="none"
+          aria-required={required}
+          as={SelectPrimitive}
+          background="none"
+          borderRadius="borderRadius20"
+          borderStyle="borderStyleSolid"
+          borderWidth="borderWidth10"
+          color="colorText"
```
BrianLincoln, line 122: More of a design question than code: does it look like it could be disabled having this lighter text?
image

BrianLincoln (approved):
Looks great!

Just the one (design) question. Not a blocker at all

packages/components/src/components/Select/Select.tsx
apvale, line 122: I think it is almost the same as what Brian asked 😅 but, 

image

Shouldn't this color be strongest when the value was selected? Having it in this way, it seems like a placeholder. 🤔 

What do you think?

packages/components/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,186 @@
...
+    backgroundColor: {
+      _: "colorBackground",
+      hover: "colorBackgroundWeakest",
+    },
+    borderColor: "colorBorderWeaker",
+  };
+};
+
+/** A select is an styled dropdown form element that allows users to select a value from a list. */
+const Select = React.forwardRef<HTMLButtonElement, SelectProps>(
```
apvale, line 87: So, we will not have a large size? From what I saw on Figma, we have two variants for size: 

image

packages/components/src/components/Select/SelectPopover.tsx
```diff
@@ -0,0 +1,45 @@
...
+import type { SelectPopoverProps as SelectPopoverPrimitiveProps } from "ariakit/select";
+import { Box } from "../../primitives/Box";
+
+export interface SelectPopoverProps
+  extends Omit<SelectPopoverPrimitiveProps, "nonce"> {
+  /** The valid contents of the listbox, i.e. SelectItem. */
+  children: NonNullable<React.ReactNode>;
+}
+
+/** A select is an styled dropdown form element that allows users to select a value from a list. */
```
apvale, line 12: What is a SelectPopover?

packages/theme/src/theme/default.ts
```diff
@@ -62,9 +62,11 @@ export const theme = {
   },
   states: {
     ...defaultTheme.states,
+    activeItem: "&[data-active-item]",
```
apvale, line 65: Nice! Didn't know that we could do this x)

packages/components/src/components/Select/SelectPopover.tsx
richbachman: @apvale SelectPopover is the dropdown with the select options.

packages/components/src/components/Select/Select.tsx
richbachman: Just missed the large size. Adding it now.

packages/components/src/components/Select/Select.tsx
richbachman: @BrianLincoln @apvale I worked with design on this a bit more and came up with a solution for an initial select option since selects don't really have placeholders. Its a `initial` boolean prop that can be added to a SelectItem. It mainly adjust styling, but could also be used for form validation down the line. We'll need to test that a bit more in an actual use-case in the app in the future.

packages/components/src/components/Select/Select.tsx
richbachman: @apvale large selects are now in storybook. Good catch.

packages/components/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,196 @@
...
+            <SelectArrow
+              as={Icon}
+              decorative
+              icon="ChevronDownIcon"
+              size={size === "large" ? "sizeIcon30" : "sizeIcon20"}
+            />
+          </Box.div>
+        </Box.button>
+        <SelectPopover state={select}>
+          {map(items, (item) => (
```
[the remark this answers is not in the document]
richbachman: @blumaa how sorting handling currently?

packages/components/src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+      <Label htmlFor={selectID}>Choose One</Label>
+      <Select hasError id={selectID} items={selectItems} name="select" />
+      <HelpText hasError id={helpTextID}>
+        This is an error message.
+      </HelpText>
+    </Box.form>
+  );
+};
+
+export const LongItem = (): JSX.Element => {
```
apvale, line 94: I think it's good having a story to test the mobile version of this one. What do you think?

apvale (approved):

packages/components/src/components/Select/SelectPopover.tsx
richbachman: Fixed that. Its different now.

packages/components/src/components/Select/Select.stories.tsx
richbachman: Yeah. Worth having that screenshot there. Its been added.

RohitRajendran (approved):

apvale (approved):

PR #326, opened 2022-12-14
feat(components): add Toast component

richbachman:
## Description of the change

This adds the Toast with success and error variants. Stories and test are included.

Screenshot 2022-12-14 at 12 16 19
Screenshot 2022-12-14 at 12 16 25
Screenshot 2022-12-14 at 12 16 33
Screenshot 2022-12-14 at 12 16 51

## Testing the change

- [ ] Load storybook

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Toast/Toast.stories.tsx
+++ b/packages/components/src/components/Toast/Toast.stories.tsx
@@ -22,7 +22,7 @@ const Template: ComponentStory<typeof Toast> = (args) => {
   const [open, setOpen] = React.useState(isChromatic() ? true : false);
 
   return (
-    <ToastProvider duration={3000}>
+    <ToastProvider duration={5000}>
       <Button onClick={() => setOpen(true)} variant="primary">
         Open toast
       </Button>
--- a/packages/components/src/components/Toast/Toast.stories.tsx
+++ b/packages/components/src/components/Toast/Toast.stories.tsx
@@ -1,6 +1,7 @@
 import type { ComponentMeta, ComponentStory } from "@storybook/react";
 import React from "react";
 import isChromatic from "chromatic/isChromatic";
+import { Anchor } from "../Anchor";
 import { Button } from "../Button";
 import { Toast } from "./Toast";
 import { ToastViewport } from "./ToastViewport";
@@ -9,6 +10,12 @@ import { ToastProvider } from ".";
 export default {
   component: Toast,
   title: "Components/Toast",
+  parameters: {
+    docs: {
+      storyDescription:
+        "ToastProvider and ToastViewport should only be used once in an application. Both should be placed in the root of the application.",
+    },
+  },
 } as ComponentMeta<typeof Toast>;
 
 const Template: ComponentStory<typeof Toast> = (args) => {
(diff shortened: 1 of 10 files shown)


packages/components/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Text } from "../../primitives/Text";
+import { Icon } from "../Icon";
+import type { ToastVariants } from "./types";
+
+export interface ToastProps
+  extends Omit<React.ComponentProps<typeof ToastPrimitive.Root>, "asChild"> {
+  /** The text used if an Anchor is used in the toast. */
+  anchorText?: string;
+  /** The href used if an Anchor is used in the toast. */
+  anchorHref?: string;
```
RohitRajendran, line 14: Can we update the type so that anchorText and anchorHref are both needed or both not needed? Right now, it's possible to provide an anchorText but no anchorHref which would result in nothing appearing.

packages/components/src/components/Toast/Toast.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+export default {
+  component: Toast,
+  title: "Components/Toast",
+} as ComponentMeta<typeof Toast>;
+
+const Template: ComponentStory<typeof Toast> = (args) => {
+  const [open, setOpen] = React.useState(isChromatic() ? true : false);
+
+  return (
+    <ToastProvider duration={5000}>
```
RohitRajendran, line 18: In actual usage, would we need to wrap each instance with the ToastProvider and ToastViewport or would there just be one that wraps the whole app? Might be good to add that info the documentation.

packages/components/src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,93 @@
...
+              lineHeight="lineHeight30"
+            >
+              {children}
+            </Text.span>
+          </ToastPrimitive.Description>
+          {anchorHref && anchorText && (
+            <>
+              {" "}
+              <ToastPrimitive.Action altText={anchorText} asChild>
+                <Anchor href={anchorHref}>{anchorText}</Anchor>
```
RohitRajendran, line 71: Are we sure this will alway be a link and might not need to call a function instead? Maybe a success toast with a undo action for example?

packages/components/src/components/Toast/Toast.stories.tsx
richbachman: `ToastProvider` and ToastViewport` will only need to be added once to the app. I'll add a comment in the stories.

blumaa (commented):
This looks good to me. And not a blocker but I was thinking would we ever need to adjust the duration? Or would we ever need to display multiple errors in one toast? It seems like one toast should be one error, right?

richbachman:
> This looks good to me. And not a blocker but I was thinking would we ever need to adjust the duration? Or would we ever need to display multiple errors in one toast? It seems like one toast should be one error, right?

@blumaa there's a `duration` prop that can be added to each Toast. The storybook docs don't reflect that. I'll be working on better docs in the future. And yeah. One error per toast.

packages/components/src/components/Toast/Toast.tsx
richbachman: I made the cta prop render anything, but for now we can only use Anchors because Button needs to be reworked so it would fit inline with text. I added a comment to the story, and I'll add a ticket to handle the Button changes.

packages/components/src/components/Toast/Toast.tsx
richbachman: Yep. I added a conditional type so the cta props are only required if `hasCta` is `true`.

RohitRajendran: Is there a need for `hasCta` now that you have the conditional type?

PR #336, opened 2022-12-16
feat(components): added multiselect ability to select

richbachman:
## Description of the change

Adjusted the base functionality so it works as a multiselect. In order to get a multiselect, an array should be passed to either `defaultValue` or `value`.

- Also removed the need for the `initial` prop on `SelectItem` by adding default text if a Select value is an empty string.!
 
Screenshot 2022-12-16 at 15 21 20

## Testing the change

- [ ] Load storybook

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Select/Select.tsx
+++ b/packages/components/src/components/Select/Select.tsx
@@ -49,6 +49,8 @@ export interface SelectProps
   items: SelectItemProps[];
   /** The `name` of the select. */
   name?: string;
+  /** The text to be used for the select placeholder. */
+  placeholder?: string;
   /** Sets the select state to required, so a user has to provide a value in order to be valid. */
   required?: boolean;
   /** Function that will be called when setting the select value state. */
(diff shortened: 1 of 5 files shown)


BrianLincoln:
Functionally this seems to be working great. Maybe a future addition, it might be nice to have badges or something to see and remove selected values. Thinking of cases like nationality where there are TONS of options it would be tough to find them. Not a great example but something like this:
image

packages/components/src/components/Select/Select.tsx
```diff
@@ -9,12 +9,16 @@ import {
...
 import type { SystemProp, Theme } from "@xstyled/styled-components";
+import isArray from "lodash/isArray";
 import { Box } from "../../primitives/Box";
+import { Text } from "../../primitives/Text";
 import { Icon } from "../Icon";
 import { SelectItem } from "./SelectItem";
 import type { SelectItemProps } from "./SelectItem";
 import { SelectPopover } from "./SelectPopover";
 
+const NO_SELECTION_STRING = "Select one";
```
BrianLincoln, line 20: Should this be override-able?

BrianLincoln (approved):
2 suggestions, not blockers. Thanks again!

richbachman:
> Functionally this seems to be working great. Maybe a future addition, it might be nice to have badges or something to see and remove selected values. Thinking of cases like nationality where there are TONS of options it would be tough to find them. Not a great example but something like this:

@BrianLincoln Yeah. I plan on that being a next iteration. I've seen examples like the one you shared, but those seem to struggle when someone has selected a lot of options. We'll probably do something where the selections live outside the select area. Here's a very unstyled example. https://codesandbox.io/s/ariakit-multi-select-combobox-with-tags-yuem9x?file=/src/App.js

packages/components/src/components/Select/Select.tsx
richbachman: Probably a good call.

richbachman: Added an optional `placeholder` prop that can replace that default string.

PR #405, opened 2023-01-12
fix(components): allow anchor and button to work as link

richbachman:
## Description of the change

Adjusted the `as` prop on both Anchor and Button so they will work better with the Link component from react router. Also added a `to` prop to Anchor for the same purpose.

Also upgraded Storybook!

## Testing the change

- [ ] Check out the new story in storybook

## Type of change

### Development

### Code review


packages/components/src/components/Anchor/Anchor.stories.tsx
```diff
@@ -24,3 +25,13 @@ External.args = {
   ...defaultArgs,
   isExternal: true,
 };
+
+export const WithReactRouter = (): JSX.Element => {
```
RohitRajendran, line 29: NextJS example too?

packages/components/src/components/Anchor/Anchor.tsx
```diff
@@ -1,12 +1,19 @@
...
+
 export interface AnchorProps
   extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "color"> {
+  /** Sets the render element of the component. */
+  as?: AnchorAsOptions;
   /** The contents of the anchor. Can be text or valid text related HTML, i.e. strong elements. */
   children: NonNullable<React.ReactNode>;
   /** Sets target to "_blank" and rel to "noreferrer noopener". */
   isExternal?: boolean;
+  /** Used with React Router or NextJS to set the route the anchor links to. */
```
RohitRajendran, line 15: NextJs Link component uses `href` prop instead of to.

packages/components/src/components/Anchor/Anchor.tsx
```diff
@@ -1,12 +1,19 @@
 import React from "react";
 import { Text } from "../../primitives/Text";
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type AnchorAsOptions<Props = any> = React.ElementType<Props>;
+
 export interface AnchorProps
   extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "color"> {
+  /** Sets the render element of the component. */
+  as?: AnchorAsOptions;
```
RohitRajendran, line 10: Should we require this if `isExternal` is false? Feels easy for people to forget otherwise and to make sure people don't include it for external links?

packages/components/src/components/Button/Button.stories.tsx
```diff
@@ -84,10 +85,14 @@ AsLink.args = {
   href: "#",
 };
 
-export const AsLabel = Template.bind({});
-AsLabel.args = {
-  as: "label",
-  children: "Label button",
+export const AsLinkWithReactRouter = (): JSX.Element => {
```
RohitRajendran, line 88: NextJS example too?

packages/components/src/components/Button/Button.tsx
```diff
@@ -36,7 +39,7 @@ export interface ButtonProps
   trailingIcon?: IconNames;
   /** Sets the button state to loading. */
   loading?: boolean;
-  /** Used with React Router or NextJS to set the route the anchor links to. */
+  /** Used with React Router or NextJS to set the route the button links to. */
```
RohitRajendran, line 42: Same thing about NextJS using href instead of to

packages/components/src/components/Anchor/Anchor.tsx
richbachman: I'm not really in favor of this in the situation someone uses Pluto for something other than Next or React-Router. Maybe they just want a plain old link.

RohitRajendran: Hmm any thoughts on how we can make sure people don't forget it then?

richbachman: We can add to the documentation. Or add tests on the app side when it happens.

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -13,7 +13,7 @@ export default {
   component: Tabs,
   subcomponents: { Tab, TabList, TabPanels, TabPanel },
   title: "Components/Tabs",
-} as ComponentMeta<typeof Tabs>;
+} as unknown as ComponentMeta<typeof Tabs>;
```
RohitRajendran, line 16: What's the deal with the unknown here?

richbachman: Since `aria-label` is required, I was getting a TS error. The unkown resets it.

richbachman: Had to actually ts-ignore this one because changing to unknown removes the story.

PR #407, opened 2023-01-12
feat(components): add Switch component and stories

richbachman:
## Description of the change

Added the Switch component stories and tests.

Screenshot 2023-01-12 at 16 37 10

Also added `colorBackgroundPrimaryWeakest` and `colorBackgroundStrongest` to design-tokens and theme.

## Testing the change

- [ ] Fire up storybook

## Type of change

### Development

### Code review


alpacatron:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ed513588c0459c8a4a74cf60ea4eee8286f0cff5
</p>

packages/components/src/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,172 @@
...
+  /** Help text for additional guidance. */
+  helpText?: React.ReactNode;
+  /** When true, indicates that the user must check the switch before the owning form can be submitted. */
+  required?: boolean;
+  /** The name of the switch. Submitted with its owning form as part of a name/value pair. */
+  name?: string;
+  /** Event handler called when the state of the switch changes. */
+  onCheckedChange?: (checked: boolean) => void;
+  /** The value given as data when submitted with a name. */
+  value?: string;
```
RohitRajendran, line 32: So we need value and checked then? I'm a little confused with the difference and how this is used.

richbachman: @RohitRajendran the value is there only if you want to pass a custom value if this lives in a form. Otherwise it functions like a checkbox. Say if you want to display content based on if a switch is checked or unchecked.

packages/components/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <Box.form onSubmit={handleSubmit(onSubmit)}>
+      <Controller
+        control={control}
+        name="enabled"
+        render={({ field }) => (
+          <Switch {...field} onCheckedChange={field.onChange}>
+            Enable forms
+          </Switch>
+        )}
+      />
```
RohitRajendran, line 83: Not for this PR but thoughts on a ControlledSwitch (or other name) that already has it wrapped in the Controller so that developers don't need to wrap it in Controller each time or more likely, create a component that's wrapped in Controller individually in the backend and talent repos?

packages/components/src/components/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,91 @@
...
+    defaultValues: { enabled: "false" },
+  });
+
+  return (
+    <Box.form onSubmit={handleSubmit(onSubmit)}>
+      <Controller
+        control={control}
+        name="enabled"
+        render={({ field }) => (
+          <Switch {...field} onCheckedChange={field.onChange}>
```
RohitRajendran, line 79: So default value for the form has the value as a string but wouldn't onCheckedChange change that into a boolean since it uses the boolean `checked` property?

RohitRajendran: I'm seeing this locally for this story 
image

RohitRajendran (approved):
🚢

packages/components/src/components/Switch/Switch.stories.tsx
richbachman: Yeah. I've been thinking about that recently. I wonder if we should just wrap our own version of react-hook-form components so they be more easily used in the different applications. Easier and consistent.

PR #440, opened 2023-01-19
feat(components): add FormSelect and FormTextarea

richbachman:
## Description of the change

Added the FormSelect and FormTextArea components, stories, and tests.

Screenshot 2023-01-19 at 16 30 08

Screenshot 2023-01-19 at 16 30 21

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/FormSelect/FormSelect.stories.tsx
+++ b/packages/components/src/components/FormSelect/FormSelect.stories.tsx
@@ -1,0 +1,92 @@
+import type { Meta } from "@storybook/react";
+import React from "react";
+import { useUID } from "react-uid";
+import type { SelectProps } from "../Select";
+import { FormSelect } from "./FormSelect";
+
+const meta: Meta<typeof FormSelect> = {
+  title: "Components/Form/FormSelect",
+  component: FormSelect,
+};
+
+export default meta;
+
+const selectItems: SelectProps["items"] = [
+  { value: "option-one", label: "Option One" },
+  { value: "option-two", label: "Option Two" },
+  { value: "option-three", label: "Option Three", disabled: true },
+  { value: "option-four", label: "Option Four" },
+  {
+    value: "option-five",
+    label:
+      "Option Five which is a really long item and should overflow to the next line",
+  },
+];
+
+export const Default = (): JSX.Element => {
+  const selectID = useUID();
+  return (
+    <FormSelect
+      id={selectID}
+      items={selectItems}
+      label="Label text"
+      name="select-field"
+    />
+  );
+};
+
+export const Required = (): JSX.Element => {
+  const selectID = useUID();
+  return (
+    <FormSelect
+      helpText="Please choose one of the values."
+      id={selectID}
+      items={selectItems}
+      label="Label text"
+      name="select-field"
+      required
+    />
+  );
+};
+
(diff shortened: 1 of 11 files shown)


packages/components/src/components/FormSelect/FormSelect.stories.tsx
```diff
@@ -0,0 +1,92 @@
+import type { Meta } from "@storybook/react";
+import React from "react";
+import { useUID } from "react-uid";
+import type { SelectProps } from "../Select";
+import { FormSelect } from "./FormSelect";
+
+const meta: Meta<typeof FormSelect> = {
+  title: "Components/Form/FormSelect",
```
blumaa, line 8: what about adding the subcomponents here? Or are we only doing it on some of the stories?

packages/components/src/components/FormSelect/FormSelect.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <FormSelect
+      defaultValue={[selectItems[0].value, selectItems[1].value]}
+      helpText="Please choose one of the values."
+      id={selectID}
+      items={selectItems}
+      label="Label text"
+      name="select-field"
+    />
+  );
+};
```
blumaa, line 92: It might be cool to add an interaction test here to see the mouse click the select. But not essential. https://storybook.js.org/docs/7.0/react/writing-tests/interaction-testing

blumaa (approved):
Seems good!

packages/components/src/components/FormSelect/FormSelect.stories.tsx
richbachman: This one really doesn't have subcomponents. Its just one that is a collection of components.

packages/components/src/components/FormSelect/FormSelect.stories.tsx
richbachman: Maybe something we could add to the Select component in the future.

RohitRajendran:
I noticed storybook complain about some accessibility issues for the FormSelect 
image

RohitRajendran (approved):
Looks good! Just the minor thing that about the excluding unnecessary types that we discussed on slack and the accessibility issues which weren't caused by this pr so could be fixed separately if wanted.

packages/components/src/components/FormSelect/FormSelect.tsx
```diff
@@ -6,7 +6,8 @@ import { HelpText } from "../HelpText";
 import type { SelectProps } from "../Select";
 import { Box } from "../../primitives/Box";
 
-export interface FormSelectProps extends Omit<SelectProps, "id"> {
+export interface FormSelectProps
+  extends Omit<SelectProps, "aria-label" | "aria-labelledby" | "id"> {
```
RohitRajendran, line 10: and aria described by

richbachman:
> accessibility issues which weren't caused by this pr so could be fixed separately if wanted.

@RohitRajendran I'll dig into it a little further, but I think Storybook is a little confused here. If I check the accessibility inspector in Chrome, it finds the button value just fine. I'm wondering if its because we're using a button with a different role, which comes from the Ariakit primitive. 🤔 

Not a blocker for this PR, but I'll create a ticket to investigate.

PR #450, opened 2023-01-24
feat(components): add ControlledFormInput for react hook form

richbachman:
## Description of the change

Added a ControlledFormInput component for easier use with React Hook Form.

Screenshot 2023-01-27 at 11 19 15

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/FormInput/ControlledFormInput.tsx
+++ b/packages/components/src/components/FormInput/ControlledFormInput.tsx
@@ -1,0 +1,70 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Controller } from "react-hook-form";
+import type { ControllerProps, Control } from "react-hook-form";
+import { FormInput } from "./FormInput";
+import type { FormInputProps } from "./FormInput";
+
+export interface ControlledFormInputProps
+  extends Omit<ControllerProps, "control" | "render">,
+    Omit<FormInputProps, "defaultValue" | "name"> {
+  /** Invoked with `useForm`. Set to any to allow `any` number of form inputs. */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  control?: Control<any>;
+  /** The text to be used for HelpText errors. */
+  errorText?: React.ReactNode;
+}
+
+const ControlledFormInput = ({
+  control,
+  disabled,
+  errorText,
+  helpText,
+  id,
+  label,
+  name,
+  placeholder,
+  readOnly,
+  required,
+  type,
+  ...props
+}: ControlledFormInputProps): JSX.Element => {
+  return (
+    <Controller
+      control={control}
+      name={name}
+      render={({ field, fieldState }) => (
+        <FormInput
+          {...field}
+          disabled={disabled}
+          hasError={fieldState.invalid}
+          helpText={fieldState.invalid ? errorText : helpText}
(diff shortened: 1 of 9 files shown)


packages/components/src/components/FormInput/ControlledFormInput.tsx
```diff
@@ -0,0 +1,70 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Controller } from "react-hook-form";
+import type { ControllerProps, Control } from "react-hook-form";
+import { FormInput } from "./FormInput";
+import type { FormInputProps } from "./FormInput";
+
+export interface ControlledFormInputProps
+  extends Omit<ControllerProps, "control" | "render">,
+    Omit<FormInputProps, "defaultValue" | "name"> {
```
RohitRajendran, line 10: Will controller pass through any of the form input props provided? I see you're spreading the remaining props into controller right now.

packages/components/src/components/FormInput/FormInput.tsx
```diff
@@ -51,6 +51,7 @@ const FormInput = React.forwardRef<HTMLInputElement, FormInputProps>(
           value={value}
           {...props}
         />
+        {!hasError && !helpText && <Box.div h="1rem" marginTop="space20" />}
```
RohitRajendran, line 54: Don't we just need to be looking at helpText and not both has error and help text?

richbachman: @RohitRajendran I think both on this one since you can have `errorText` without having `helpText`.

packages/components/src/components/FormInput/ControlledFormInput.tsx
richbachman: Right. The props need to spread to the Controller, but I should pull out any FormInput props and spread those on the actual FormInput. Good catch.

richbachman: @RohitRajendran after digging into this more, I ended up using the necessary props for Controller, and then spread everything else to the input. That way we catch any extra props that need to be added directly to that. Controller has all that it needs.

packages/components/src/components/FormInput/FormInput.tsx
RohitRajendran: error text would come in through the`helpText` prop though right? So If there's HelpText then we don't need the spacer.

richbachman: @RohitRajendran I added an `errorText` prop as well so teams can pass different strings.

RohitRajendran: That errorText prop was in ControlledFormInput and not in the FormInput though. I see in ControlledFormInput that it still gets passed into FormInput through the helpText prop so you still wouldn't need to check `hasError` inside of this component, just `helpText`. Or were you planning on adding the errorText prop to FormInput and HelpText components too?

richbachman: I see what you're saying now. Yep. Just checking `helpText`. `errorText` shouldn't be passed to FormInput.

blumaa (approved):
Looks good. Couldn't find anything to add/edit.

PR #679, opened 2023-03-24
fix: use Box.div on ProgressBar fixing tokens usage

rssilva:
## Description of the change
- Change the ProgressBar component to use Box.div instead of Styled components so the token overwriting can be correct

## Testing the change

- [x] Write your testing instructions here

- you can edit `.storybook/preview.js` rewriting tokens like below. This was not affecting the component before and now it works properly

```typescript
      <ThemeProvider
        theme={{
          ...theme,
          colors: {
            ...theme.colors,
            colorAvatarBackgroundGreen: "limegreen",
            colorAvatarBackgroundPink: "purple",
          },
        }}
      >
```

and change stories snippet like

```typescript
export const WithCustomColors = (): JSX.Element => (
  <Box.div display="flex" flexDirection="column" gap="space40">
    <ProgressBar
      backgroundColor="colorAvatarBackgroundPink"
      indicatorColor="colorAvatarBackgroundGreen"
      value={25}
    />
    <Box.div backgroundColor={"colorAvatarBackgroundGreen"}>Hello!</Box.div>
    <Box.div backgroundColor={"colorAvatarBackgroundPink"}>Good bye!</Box.div>
  </Box.div>
);
```

before this change:

Screen Shot 2023-03-24 at 12 23 43

after this change:

Screen Shot 2023-03-24 at 12 24 38

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/ProgressBar/ProgressBar.tsx
+++ b/packages/components/src/components/ProgressBar/ProgressBar.tsx
@@ -19,7 +19,13 @@ export interface ProgressBarProps {
 /** Displays an indicator showing the completion progress of a task, typically displayed as a progress bar. */
 const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
   (
-    { value, size = "small", indicatorColor, backgroundColor, ...props },
+    {
+      value,
+      size = "small",
+      indicatorColor = "colorBackgroundPrimaryWeak",
+      backgroundColor = "colorBackgroundWeak",
+      ...props
+    },
     ref
   ) => {
     const borderRadius = size === "large" ? "borderRadius50" : "borderRadius10";
@@ -27,7 +33,7 @@ const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
     return (
       <Box.div
         as={Root}
-        backgroundColor={backgroundColor || "colorBackgroundWeak"}
+        backgroundColor={backgroundColor}
         borderRadius={borderRadius}
         h={size === "large" ? "0.75rem" : "0.4rem"}
         overflow="hidden"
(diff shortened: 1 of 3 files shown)


packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -1,8 +1,7 @@
 import React from "react";
-import { styled, theme } from "@localyze-pluto/theme";
```
rssilva, line 2: By importing theme.colors from this package it would never be able to get the tokens that are overwritten on the ThemeProvider

packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -67,22 +25,27 @@ const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
     const borderRadius = size === "large" ? "borderRadius50" : "borderRadius10";
 
     return (
-      <StyledProgress
+      <Box.div
         as={Root}
-        background={backgroundColor}
-        h={size === "large" ? "space40" : "space25"}
-        radius={borderRadius}
+        backgroundColor={backgroundColor || "colorBackgroundWeak"}
```
RohitRajendran, line 30: Can we move the default into the params similar to `size`? The stories don't seem to be set up properly (or just broken) for it to show right now but storybook will then show the default value in the props table for the component. Also, just makes it easier to know what the default values for props are.

```    { value, size = "small", indicatorColor = "colorBackgroundPrimaryWeak", backgroundColor = "colorBackgroundWeak", ...props },
```

packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -67,22 +25,27 @@ const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
...
         ref={ref}
         value={value}
         {...props}
       >
-        <StyledIndicator
+        <Box.div
           as={Indicator}
-          background={indicatorColor}
-          radius={borderRadius}
+          backgroundColor={indicatorColor || "colorBackgroundPrimaryWeak"}
```
RohitRajendran, line 41: Same thing with this.

RohitRajendran (commented):
Thanks for fixing this! Just one suggestion.

packages/components/src/components/ProgressBar/ProgressBar.tsx
rssilva: oh good call! Just changed!

RohitRajendran (approved):

PR #682, opened 2023-03-27
fix(components): fix style props in progressBar

blumaa:
## Description of the change

We need to be able to style the indicator and the background of the indicator, so we added props for both of these so we could over write them.

Screenshot 2023-03-27 at 16 00 38

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/ProgressBar/ProgressBar.stories.tsx
+++ b/packages/components/src/components/ProgressBar/ProgressBar.stories.tsx
@@ -30,6 +30,22 @@ export const WithCustomColors = (): JSX.Element => (
   </Box.div>
 );
 
+export const WithOverWrittenStyles = (): JSX.Element => (
+  <Box.div display="flex" flexDirection="column" gap="space40">
+    <ProgressBar
+      backgroundColor="colorAvatarBackgroundPink"
+      indicatorColor="colorAvatarBackgroundGreen"
+      indicatorStyle={{
+        backgroundColor: "purple",
+      }}
+      style={{
+        backgroundColor: "red",
+      }}
+      value={25}
+    />
+  </Box.div>
+);
+
 export const Animated = (): JSX.Element => {
   const [progress, setProgress] = useState(13);
 
--- a/packages/components/src/components/ProgressBar/ProgressBar.tsx
+++ b/packages/components/src/components/ProgressBar/ProgressBar.tsx
@@ -14,6 +14,10 @@ export interface ProgressBarProps {
   backgroundColor?: SystemProp<keyof Theme["colors"], Theme>;
   /** Sets the size of the progress bar. */
   size?: ProgressBarSizeOptions;
+  /** The style of the progress indicator background color, should be an object with html style attributes */
+  style?: React.CSSProperties;
+  /** The style of the progress indicator, should be an object with html style attributes */
+  indicatorStyles?: React.CSSProperties;
 }
 
 /** Displays an indicator showing the completion progress of a task, typically displayed as a progress bar. */
@@ -24,6 +28,7 @@ const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
       size = "small",
       indicatorColor = "colorBackgroundPrimaryWeak",
       backgroundColor = "colorBackgroundWeak",
+      indicatorStyles,
       ...props
     },
     ref
@@ -47,7 +52,10 @@ const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
           backgroundColor={indicatorColor}
           borderRadius={borderRadius}
           h="100%"
-          style={{ transform: `translateX(-${100 - value}%)` }}
+          style={{
+            transform: `translateX(-${100 - value}%)`,
+            ...indicatorStyles,
+          }}
           transition="transform 660ms cubic-bezier(0.65, 0, 0.35, 1)"
           w="100%"
         />
--- a/.changeset/violet-poems-remain.md
+++ b/.changeset/violet-poems-remain.md
@@ -1,0 +1,5 @@
+---
+"@localyze-pluto/components": patch
+---
+
+Added style props to the progressBar


packages/components/src/components/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -24,7 +24,12 @@ export const WithCustomColors = (): JSX.Element => (
   <Box.div display="flex" flexDirection="column" gap="space40">
     <ProgressBar
       backgroundColor="colorAvatarBackgroundPink"
-      indicatorColor="colorAvatarBackgroundGreen"
```
rssilva, line 27: I believe that is not necessary or we can add a new story like "overwriting styles"

packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -14,6 +14,10 @@ export interface ProgressBarProps {
   backgroundColor?: SystemProp<keyof Theme["colors"], Theme>;
   /** Sets the size of the progress bar. */
   size?: ProgressBarSizeOptions;
+  /** The style of the progress indicator background color, should be an object with html style attributes */
+  style?: React.CSSProperties;
+  /** The style of the progress indicator, should be an object with html style attributes */
+  indicatorStyle?: React.CSSProperties;
```
apvale, line 20: Why isn't the already present prop `indicatorColor` enough for this? 🤔

packages/components/src/components/ProgressBar/ProgressBar.stories.tsx
apvale: Yes, it's better to create a new story for this case.

packages/components/src/components/ProgressBar/ProgressBar.tsx
apvale, line 20:
```suggestion
  indicatorStyles?: React.CSSProperties;
```

packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -14,6 +14,10 @@ export interface ProgressBarProps {
   backgroundColor?: SystemProp<keyof Theme["colors"], Theme>;
   /** Sets the size of the progress bar. */
   size?: ProgressBarSizeOptions;
+  /** The style of the progress indicator background color, should be an object with html style attributes */
+  style?: React.CSSProperties;
```
apvale, line 18: This prop is not being used in the component, right?

apvale: By having this:
```
<ProgressBar
      backgroundColor="colorAvatarBackgroundPink"
      indicatorStyles={{
        backgroundColor: "purple",
      }}
      value={25}
    />
```
I was able to see this: 

image

The backgroundColor of the indicator is the style that you are already adding as prop.

blumaa: Actually there is a backgroundColor for the progress already made and backgroundColor for the progress not done yet.

rssilva: I don't know what is the deal here but TS is complaining 😞 

Screen Shot 2023-03-27 at 12 31 38

blumaa: Yes but Ana now try overwriting the pink background color

nyan07: Why are we adding an `style` property?

rssilva: to be possible to overwrite styles from a token with any custom ones

packages/components/src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -24,6 +28,7 @@ const ProgressBar = React.forwardRef<HTMLDivElement, ProgressBarProps>(
       size = "small",
       indicatorColor = "colorBackgroundPrimaryWeak",
       backgroundColor = "colorBackgroundWeak",
+      indicatorStyle,
```
nyan07, line 31: Why not just create a token `colorBackgroundIndicator` for example, replace the default token on line 29, and override it on the talent app?

packages/components/src/components/ProgressBar/ProgressBar.tsx
nyan07: It would solve the problem, right? https://github.com/Localitos/pluto/pull/682#discussion_r1149447921

packages/components/src/components/ProgressBar/ProgressBar.tsx
rssilva: We could do but I don't believe that `colorBackgroundIndicator` is a good variable according with many directives

 **"Design tokens are name and value pairings that represent small, repeatable design decisions."**

This is not a decision, this is a more flexible and customizable component

packages/components/src/components/ProgressBar/ProgressBar.tsx
nyan07: But it will create a mess on the talent app. Because we will need to replace colors in all components. Besides that it will also create a mess on the DS. All components will be too open to adding any styles, we can easily lose track of what we want for the DS, and if a token changes, we will need to figure out every place where styles were added. 

If you use tokes to defined it You just need to map on the talent app:

```typescript
{
  tokenX : colorPrimaryFromBackend
}
```

nyan07: This means that the user can add the color, they don't need to add the token. The BR will store the colors and the FE will map wich color should be added to which token.

packages/components/src/components/ProgressBar/ProgressBar.tsx
nyan07: Yes, I agree. But this is happening just because the MVP of the theming project wants to change only small stuff. The DS should not be affected by this decision. And the maintenance of the talent app would be strongly affected by it. 

Honestly, I am not sure which one is the worse option. But we need to figure out a better way to do it.

packages/components/src/components/ProgressBar/ProgressBar.tsx
nyan07: We can add the tokens you need on Pluto instead of adding `styles` tags in all components we need to change. But my suggestion is not to create tokens for specific components but to map what colors you need to change. For example:

You need to update the backgroundColor from `<ProgressBar />`  and `<Button />` and currently they use the token `colorBackground`. 

But the `<Box.div backgroundColor="colorBackground">` uses the same token. And you don't want to change it. 

My suggestion: create a new token, for example `colorBackgroundPrimaryComponents` and use it in all components you want to apply the theming.

Of course,  `colorBackground` and `colorBackgroundPrimaryComponents`  would have the same token on Pluto, but you would be able to override only the one you need.

rssilva:
The team decided not allowing style overwrite in this component so this PR shouldn't proceed with its modifications!

PR #691, opened 2023-03-28
feat: add theme related tokens

rssilva:
## Description of the change
- Add tokens that will be used to change the themes at specific parts of applications

## Testing the change

## Type of change

### Development

### Code review


packages/design-tokens/src/tokens/color.tokens.json
```diff
@@ -157,6 +163,51 @@
...
+    "steps-text": {
+      "value": "#FFFFFF"
+    },
+    "steps-map-background": {
+      "value": "#102EE9"
+    },
+    "footer-text": {
+      "value": "#041162"
+    },
+    "footer-secondary-text": {
```
blumaa, line 187: It seems like in Pluto the convention is type-feature-variant

```suggestion
    "text-footer-secondary": {
```

packages/design-tokens/src/tokens/color.tokens.json
```diff
@@ -157,6 +163,51 @@
...
+    "header-text-inverse": {
+      "value": "#FFFFFF"
+    },
+    "header-decoration": {
+      "value": "#6EE7B7"
+    },
+    "map-icon": {
+      "value": "#4A63FC"
+    },
+    "steps-text": {
```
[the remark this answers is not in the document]
FranziGrieser: just throwing in another one here: 
```suggestion
    "step-container-text": {
```

or 

```suggestion
    "step-component-text": {
```

packages/design-tokens/src/tokens/color.tokens.json
FranziGrieser: I think it would be good to start with the element like `footer` here

packages/design-tokens/src/tokens/color.tokens.json
```diff
@@ -157,6 +163,51 @@
...
+    "footer-text": {
+      "value": "#041162"
+    },
+    "footer-secondary-text": {
+      "value": "#0F172A"
+    },
+    "footer-background": {
+      "value": "#C8F0EF"
+    },
+    "primary-button-text": {
```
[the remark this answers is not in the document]
FranziGrieser: and the following ones `button-background-primary`, `button-background-hover-primary`and so on. What do you think about that?

I think most of them follow that guideline by now.

packages/design-tokens/src/tokens/color.tokens.json
```diff
@@ -157,6 +163,51 @@
...
+    "header-decoration": {
+      "value": "#6EE7B7"
+    },
+    "map-icon": {
+      "value": "#4A63FC"
+    },
+    "steps-text": {
+      "value": "#FFFFFF"
+    },
+    "steps-map-background": {
```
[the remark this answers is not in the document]
rssilva: Actually this is changing to be only map-background because this component is used outside of stepProgress too

rssilva:
@nyan07 @FranziGrieser @blumaa after reading this I did a 

```
component
  property
    variation
      state
```

packages/theme/src/__tests__/__snapshots__/theme.test.tsx.snap
```diff
@@ -69,6 +75,17 @@ Object {
...
+    "colorMapIcon": "#4A63FC",
+    "colorPrimary": "#102EE9",
+    "colorPrimaryButtonBackground": "#102EE9",
+    "colorPrimaryButtonBackgroundHover": "#041162",
+    "colorPrimaryButtonBackgroundOutlineFocus": "#102EE9",
+    "colorPrimaryButtonText": "#FFFFFF",
+    "colorProgressBarBackgroundPrimary": "#FFFFFF",
+    "colorProgressBarIndicatorPrimary": "#102EE9",
+    "colorSecondary": "#76FAC1",
+    "colorStepsMapBackground": "#102EE9",
```
FranziGrieser, line 87: so maybe also remove steps here? 🤔 
```suggestion
    "colorMapBackground": "#102EE9",
```

packages/theme/src/__tests__/__snapshots__/theme.test.tsx.snap
```diff
@@ -69,6 +75,17 @@ Object {
...
+    "colorStepsText": "#FFFFFF",
```
FranziGrieser, line 88: and should this then be 
```suggestion
    "colorStepProgressText": "#FFFFFF",
```
?

FranziGrieser (commented):
Thanks for the changes, makes sense to name them in a way you did.

Maybe it would be worth going through the other tokens to adapt the naming there as well (not in this PR though)

packages/theme/src/__tests__/__snapshots__/theme.test.tsx.snap
rssilva: hmmm this is just a test snapshot and should reflect the json 🤔. Something is off!

packages/design-tokens/src/tokens/color.tokens.json
```diff
@@ -157,6 +157,57 @@
     "icon-stronger": {
       "value": "#0F172A"
     },
+    "custom-primary": {
```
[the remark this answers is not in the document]
rssilva: Indeed! 💯

packages/theme/src/__tests__/__snapshots__/theme.test.tsx.snap
```diff
@@ -61,6 +61,23 @@ Object {
     "colorBorderWarning": "#D97706",
     "colorBorderWeaker": "#CBD5E1",
     "colorBorderWeakest": "#E2E8F0",
+    "colorCustomButtonBackgroundPrimary": "#102EE9",
```
nyan07, line 64: Do we need the custom prefix? If not, I would prefer to remove it.
We will use `colorCustomButtonBackgroundPrimary` for all primary buttons, right? So, maybe we can call it just  `colorButtonBackgroundPrimary`.

blumaa (approved):
This is looking good to me!

packages/theme/src/__tests__/__snapshots__/theme.test.tsx.snap
nyan07: I don't know if we have exceptions, like using the `colorCustomSecondary` and `colorSecondary`. If yes, maybe we can think about something for the `colorCustomSecondary`.

PR #702, opened 2023-03-29
feat: add border prop to avatar and borderWidth tokens

apvale:
## Description of the change

We need to have the orange border in Avatars to render it as a premium user in our internal apps. We also had this prop in Figma, but we didn't have it here. 

This PR adds:

* New borderWidth tokens
* New prop to Avatar component based on PremiumUser

image

## Testing the change

## Type of change

### Development

### Code review


packages/components/src/components/Avatar/Avatar.tsx
```diff
@@ -210,6 +257,7 @@ const Avatar = React.forwardRef<HTMLDivElement, AvatarProps>(
           overflow="hidden"
           ref={ref}
           {...ariaProps}
+          {...(premiumUser && getBorderStyles(size))}
```
blumaa, line 260: Why the name here `premiumUser`? Will there be other cases to use border style for the avatar? It seems like pluto shouldn't care about a user. What about `withBorderStyles` instead?

nyan07: Design systems usually contain five levels: atoms, molecules, organisms, templates, and pages. This means that we can have business-related stuff on it. 

image

https://atomicdesign.bradfrost.com/chapter-2/

But in this specific case, we are talking about an atom, so it would probably be better to use `withBorder: boolean`.

blumaa: I'm excited to read this when I get time! thx Dani

apvale: @nyan07 In this case, it will always be an orange border. Should we also accept borderColor in the Avatar component? Should this be this flexible? 🤔 

So, changing the `withBorder` to an object with `{ value: "", borderColor: "" }`.

apvale: But if we decide not to use `borders` to identify users? 🤷🏻‍♀️ We will need to update the component again just to start using different style properties and adding a new `withOtherStyleType`. By using `premiumUser` or `userType`, we will need just to change the component internally.

blumaa: It's not exactly the same but somehow it reminds me of this: 

`// ❌ Does not indicate that `labelTooltip` depends on `label`!
type Props = {
  label?: string;
  labelTooltip?: string;

  /* ... */
}

// ✅ `tooltip` cannot exist without `text`!
type Props = {
  label?: string | { text: string; tooltip?: string };
}

export TextInput: React.FC<Props> = (props) => {
  const label = props.label ? normalizeLabel(props.label) : null;

 /* ... */
};`

from https://www.gabe.pizza/notes-on-component-libraries/

nyan07: It is also a good point. We have two ways to get around it (no matter the property name).

a) `border: boolean`: that in future could be `border: boolean | { some object }`
b) dispatch an expectation when borderColor is set without `withBorder`.

apvale: I changed it here https://github.com/Localitos/pluto/pull/702/commits/36885b126d5f306873f79784762465decfee0e1e
Can you check it again, please? 🙏🏻

blumaa (approved):
lgtm Nicely done Ana!

nyan07 (approved):

PR #710, opened 2023-03-30
Controlled component fixes

RohitRajendran:
## Description of the change

The controlled version of the form components were throwing an error when using from the package. I believe this might be because react hook forms is not listed as a dependency even though these components really on an import from that package similar to the other dependencies like `@heroicons/react`. Since `yarn link` doesn't work, I will need this to get published to test it out.

I've also updated those controlled components to omit `hasError` from the props since that doesn't get used and to automatically use an error message from the field state so that we don't have to pass it in.

## Testing the change

- [ ] Try using the ControlledFormX components with react hook forms

## Type of change


packages/components/src/components/FormInput/FormInput.stories.tsx
```diff
@@ -12,9 +12,6 @@ import { ControlledFormInput } from "./ControlledFormInput";
 const meta: Meta<typeof FormInput> = {
   title: "Components/Form/FormInput",
   component: FormInput,
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore Table has a required prop which confuses Storybook here.
-  subcomponents: { ControlledFormInput },
 };
```
RohitRajendran, line 15: This was causing an error on the docs page so I've removed it. There seems to be a separate issue where the props table doesn't seem to be working in our storybook that we should look into.

packages/components/package.json
```diff
@@ -35,7 +35,8 @@
     "@radix-ui/react-switch": "^1.0.1",
     "@radix-ui/react-toast": "^1.1.2",
     "ariakit": "^2.0.0-next.43",
-    "react-dropzone": "^14.2.3"
+    "react-dropzone": "^14.2.3",
+    "react-hook-form": "^7.43.2"
```
RohitRajendran, line 39: I added this here because the other components are listed here but do we really need to be listing the package twice? I thought just having it in `peerDepdencies` would have been enough.

I'm planning on going with this for now to unblock my work since we can't quickly test these changes without publishing the packages. Then we can come back and test this a little more to see if this is actually needed later.

blumaa (approved):
I can't see any probs but maybe best to ask someone else also! Nice

jvaladas (approved):
Looks good to me 👍  
Left a question

packages/components/src/components/FormTextArea/FormTextArea.stories.tsx
```diff
@@ -123,18 +117,18 @@ export const AsControlledFormTextArea = (): JSX.Element => {
   const onSubmit: SubmitHandler<FormInputs> = (data) =>
     alert(JSON.stringify(data, null, 2));
 
+  useEffect(() => {
+    // Trigger validation on mount
+    trigger();
```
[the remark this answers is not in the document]
RohitRajendran: That's if you just want to trigger validation on that one field. In this case, I'm just forcing a validation on all the fields  after mount.

PR #736, opened 2023-04-04
fix: make file upload title disabled

apvale:
## Description of the change

We should also handle the file uploaded title when the component status is disabled. For this case, we are sending the status to the FileUploaderTitle and checking if the status is disabled to create the link or not in the title.

image

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/.changeset/green-sloths-wait.md
+++ b/.changeset/green-sloths-wait.md
@@ -1,0 +1,5 @@
+---
+"@localyze-pluto/components": patch
+---
+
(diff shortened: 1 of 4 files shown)


nyan07:
Do we have a test to cover this behavior?

apvale:
> Do we have a test to cover this behavior?

The test should be story we already have for the disabled scenario in a FileUploader. Do you think I should create another story for this?

nyan07 (approved):

PR #740, opened 2023-04-04
feat: create controlled checkbox

RohitRajendran:
## Description of the change

This creates a ControlledCheckbox component to be easily used with React Hook Forms.

## Testing the change

- [ ] Visit the As Controlled Checkbox story for Checkbox

## Type of change

### DIFF
--- a/packages/components/src/components/Checkbox/ControlledCheckbox/ControlledCheckbox.tsx
+++ b/packages/components/src/components/Checkbox/ControlledCheckbox/ControlledCheckbox.tsx
@@ -1,0 +1,55 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Control, Controller } from "react-hook-form";
+import type { ControllerProps } from "react-hook-form";
+import { CheckedState } from "@radix-ui/react-checkbox";
+import { Checkbox } from "../Checkbox";
+import type { CheckboxProps } from "../Checkbox";
+
+export interface ControlledCheckboxProps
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Checkbox/ControlledCheckbox/ControlledCheckbox.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { CheckedState } from "@radix-ui/react-checkbox";
+import { Checkbox } from "../Checkbox";
+import type { CheckboxProps } from "../Checkbox";
+
+export interface ControlledCheckboxProps
+  extends Pick<ControllerProps, "name" | "rules">,
+    Omit<CheckboxProps, "checked" | "error" | "name" | "onBlur" | "value"> {
+  // Invoked with `useForm`. Set to any to allow `any` number of form inputs.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  control?: Control<any>;
```
rssilva, line 14: out of curiosity: can/should be `unknown` here?

rssilva (approved):
lgtm! Just a small question regarding types more curiosity than anything

packages/components/src/components/Checkbox/ControlledCheckbox/ControlledCheckbox.tsx
RohitRajendran: It doesn't like being set to unknown 😞 
image

PR #749, opened 2023-04-06
Select component fixes

RohitRajendran:
## Description of the change

The onBlur of the Select was getting triggered when clicking on a menu item causing issues with validating on blur. The Select now follows a similar approach seen in AriaKit's FormSelect component to prevent this.

I've also made `control` prop required for the `ControlledFormSelect` component as the component throws an error without it.

## Testing the change

## Type of change

### DIFF
--- a/packages/components/src/components/Select/Select.tsx
+++ b/packages/components/src/components/Select/Select.tsx
@@ -192,6 +192,16 @@ const Select = React.forwardRef<HTMLButtonElement, SelectProps>(
           w="100%"
           {...getSelectStyles(disabled, hasError)}
           {...props}
+          onBlur={(event) => {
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Select/Select.tsx
```diff
@@ -192,6 +192,16 @@ const Select = React.forwardRef<HTMLButtonElement, SelectProps>(
           w="100%"
           {...getSelectStyles(disabled, hasError)}
           {...props}
+          onBlur={(event) => {
+            if (event.defaultPrevented) {
```
BrianLincoln, line 196: TIL

BrianLincoln (approved):
Looks good, thanks for figuring this out

PR #751, opened 2023-04-11
feat(tokens): add color background outline hover token

rssilva:
## Description of the change
- Add a token to change the outline button background color on hover

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/design-tokens/src/tokens/color.tokens.json
+++ b/packages/design-tokens/src/tokens/color.tokens.json
@@ -79,6 +79,9 @@
     "background-decorative-weak": {
       "value": "#FFF1DB"
     },
+    "background-outline-hover": {
+      "value": "#EBEDFF"
+    },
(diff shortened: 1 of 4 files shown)


packages/design-tokens/src/tokens/color.tokens.json
```diff
@@ -79,6 +79,9 @@
     "background-decorative-weak": {
       "value": "#FFF1DB"
     },
+    "background-outline-hover": {
```
blumaa, line 82: We don't need a background-outline token also?

blumaa (approved):
lgtm. Left one small comment but no need to respond.

packages/design-tokens/src/tokens/color.tokens.json
rssilva: I believe that we are good because this will solve the contrast issue!

PR #761, opened 2023-04-12
Fix button stylings getting overwritten

RohitRajendran:
## Description of the change

The Button as Link's color and text decoration was getting overwritten on hover by other conflicting design systems. This makes those styles explicit so that this doesn't happen.

Issue:
image

## Testing the change

- [ ] Check the Button story to make sure styles still look right

## Type of change


BrianLincoln (approved):
Nice!

PR #763, opened 2023-04-18
fix: show upload button in case of error

nyan07:
## Description of the change
In case of error the component was showing the trash icon instead of the upload button.
image

## Type of change

- Bug fix (non-breaking change that fixes an issue)

### Development

### Code review

### DIFF
--- a/.changeset/serious-avocados-invent.md
+++ b/.changeset/serious-avocados-invent.md
@@ -1,0 +1,5 @@
+---
+"@localyze-pluto/components": patch
+---
+
+[FileUploader] Fix: Remove IconTrash button when file uploader has an error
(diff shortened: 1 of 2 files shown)


alpacatron:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against b7d30cd269d7b40921c10ccde326efc62569dbd7
</p>

RohitRajendran:
Does this still work if there's an issue with an uploaded file?

nyan07:
> > Does this still work if there's an issue with an uploaded file?
> 
> What do you mean? thinking
> 
> I think the PR mostly removes the IconTrash when it shouldn't be shown. But it doesn't add the Upload button, since we need to add it manually. cc @nyan07
> 
> But now talking about this, I didn't know, but we have a FileUploaderButton in Pluto, and we never used it. In the talent app, we use this one https://github.com/Localitos/talent-web/blob/main/src/modules/steps/components/uploadDocument/UploadDocument.tsx#L92. eyes

The button was added before the uppy (modal) integration. It is used for uploading without libraries (as we were doing before).

PR #766, opened 2023-04-18
Revert "fix: show upload button in case of error"

apvale:
Reverts Localitos/pluto#763

* The changes were made to validation errors. But we realized that the current behavior is only specific for uploading errors. So, we need to discuss before with the Designers if we want the same view for both cases.

### DIFF
--- a/.changeset/serious-avocados-invent.md
+++ b/.changeset/serious-avocados-invent.md
@@ -1,5 +1,0 @@
----
-"@localyze-pluto/components": patch
----
-
-[FileUploader] Fix: Remove IconTrash button when file uploader has an error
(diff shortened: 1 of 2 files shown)


alpacatron:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5eafb56fb6bf63ecbfc1dd53c06594a33afe34c6
</p>

PR #781, opened 2023-04-24
fix: only render select items when open

RohitRajendran:
## Description of the change

We were rendering all the select items even when the menu isn't open which causes performance issues when there's a lot of items.

This only renders the select items when the popover is open to help with this.

## Testing the change

- [ ] Check out the Select stories to make sure everything still works as expected.

## Type of change


rssilva (approved):
✨

PR #793, opened 2023-04-27
chore: change File Uploader error handling to show buttons correctly

rssilva:
## Description of the change
- Handle errors on FileUploader to not show a remove button if there is no file yet
- Add a story with required field to demonstrate it

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/FileUploader/FileUploader.stories.tsx
+++ b/packages/components/src/components/FileUploader/FileUploader.stories.tsx
@@ -259,7 +259,9 @@ export const WithRequiredButton = (): React.ReactElement => {
   const [file, setFile] = useState<File | undefined>();
   const [error, setError] = useState("");
 
-  const onSubmit = (ev: React.MouseEvent<HTMLButtonElement>) => {
+  const onSubmit = (
+    ev: React.FormEvent<HTMLFormElement> | React.MouseEvent<HTMLButtonElement>
+  ) => {
     ev.preventDefault();
 
     if (file) {
@@ -272,7 +274,7 @@ export const WithRequiredButton = (): React.ReactElement => {
   };
 
   return (
-    <form>
+    <form onSubmit={onSubmit}>
       <FileUploader errorMessage={error} fileName={file?.name} label="Passport">
         <FileUploaderButton
           id="upload-passport"
--- a/packages/components/src/components/FileUploader/FileUploader.stories.tsx
+++ b/packages/components/src/components/FileUploader/FileUploader.stories.tsx
@@ -254,3 +254,49 @@ export const HandlingSubmitErrors = (): React.ReactElement => {
     </form>
   );
 };
+
+export const WithRequiredButton = (): React.ReactElement => {
+  const [file, setFile] = useState<File | undefined>();
+  const [error, setError] = useState("");
+
+  const onSubmit = (ev: React.MouseEvent<HTMLButtonElement>) => {
+    ev.preventDefault();
+
+    if (file) {
+      setError("");
+      setFile(file);
+    } else {
+      setError("Please upload a file");
+      setFile(undefined);
+    }
+  };
+
+  return (
+    <form>
+      <FileUploader errorMessage={error} fileName={file?.name} label="Passport">
+        <FileUploaderButton
+          id="upload-passport"
+          onChange={(ev) => {
+            setFile(ev.target.files?.[0]);
+            setError("");
+          }}
+          required
+          type="button"
+          variant={"secondary"}
+        >
+          Upload
+        </FileUploaderButton>
+      </FileUploader>
+      <Button
+        onClick={onSubmit}
+        style={{
+          marginTop: "30px",
+        }}
+        type="submit"
+        variant="primary"
+      >
+        Submit
+      </Button>
+    </form>
+  );
+};
--- a/packages/components/src/components/FileUploader/FileUploader.stories.tsx
+++ b/packages/components/src/components/FileUploader/FileUploader.stories.tsx
@@ -211,7 +211,7 @@ export const WithUppy = (): React.ReactElement => {
   );
 };
 
-export const AsRequiredFileUploader = (): React.ReactElement => {
+export const HandlingSubmitErrors = (): React.ReactElement => {
   const [file, setFile] = useState<File | undefined>();
   const [error, setError] = useState("");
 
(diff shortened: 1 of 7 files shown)


packages/components/src/components/FileUploader/FileUploader.tsx
```diff
@@ -46,6 +46,9 @@ export interface FileUploaderProps
 
   /** Disables the file uploader */
   disabled?: boolean;
+
+  /** When `true`, indicates that the user must upload a file before the form can be submitted. */
+  required?: boolean;
```
RohitRajendran, line 51: Do we need the required prop here? Don't we actually need it in the FileUploaderButton so that it gets applied to the hidden input?

packages/components/src/components/FileUploader/utils/getStatus.ts
```diff
@@ -5,13 +5,16 @@ export const getStatus = ({
...
 }: {
   progress: number;
   fileUrl?: string;
   errorMessage?: string;
+  required?: boolean;
 }): FileUploaderStatus => {
   if (progress > 0 && progress < 100) return "loading";
   if (!isEmpty(fileUrl)) return "success";
   if (!isEmpty(errorMessage)) return "error";
+  if (!isEmpty(errorMessage) && required && !fileUrl) return "error";
```
RohitRajendran, line 18: This seems unnecessary. It would never actually get to this line because it would always return at line 17 if there's an error message.  I don't think we really need this either, if there's an error message then it should just be an error state, we don't need to check the other properties here.

packages/components/src/components/FileUploader/FileUploader.tsx
```diff
@@ -63,13 +66,19 @@ const FileUploader = React.forwardRef<HTMLDivElement, FileUploaderProps>(
...
     },
     ref
   ) => {
     const isMobile = useDown("md");
     const status = getStatus({ progress, fileUrl, errorMessage });
     const isMobileUploading = isMobile && status === "loading";
     const isUploadedWithoutName = status === "success" && !fileName;
+    const hasFile = fileName || fileUrl;
+    const shouldShowChildren =
+      status === "waiting" || (required && errorMessage);
```
RohitRajendran, line 79: Do we need to be checking `required` and `errorMessage` here? Just checking if it's an error state and that hasFile is false would be more comprehensive.

packages/components/src/components/FileUploader/utils/getStatus.ts
rssilva: jesus 🤦

packages/components/src/components/FileUploader/FileUploader.tsx
rssilva: There are some places that are not using the <FileUploaderButton /> together with FileUploader. Maybe we should have both options?

packages/components/src/components/FileUploader/FileUploader.tsx
```diff
@@ -63,13 +66,18 @@ const FileUploader = React.forwardRef<HTMLDivElement, FileUploaderProps>(
...
+      required,
     },
     ref
   ) => {
     const isMobile = useDown("md");
     const status = getStatus({ progress, fileUrl, errorMessage });
     const isMobileUploading = isMobile && status === "loading";
     const isUploadedWithoutName = status === "success" && !fileName;
+    const hasFile = fileName || fileUrl;
+    const shouldShowChildren = status === "waiting" || (required && !hasFile);
```
RohitRajendran, line 78: If there's an error but there's no file, then it wouldn't show children or the remove button
```suggestion
    const shouldShowChildren = status === "waiting" || (status === "error" && !hasFile);
```

rssilva: oh right! In this case we wouldn't need the `required` prop 🤔

RohitRajendran: Yeah, I don't think the FileUploader component really needs to care about `required`. It just needs to be able to show the errors properly.

packages/components/src/components/FileUploader/FileUploader.test.tsx
```diff
@@ -127,6 +127,39 @@ describe("<FileUploader />", () => {
...
+
+      expect(screen.getByText("Permit")).toBeInTheDocument();
+      expect(
+        screen.getByRole("button", { name: "Upload" })
+      ).toBeInTheDocument();
+      expect(screen.queryByRole("alert")).not.toBeInTheDocument();
+    });
+  });
+
   it('renders correctly when state is "error"', () => {
```
RohitRajendran, line 163: Nit: Shouldn't this be part of the previous describe? Can we also update the test naming to reflect the difference in `fileName` being provided on whether it shows the upload button or the remove button?

packages/components/src/components/FileUploader/utils/getStatus.test.ts
```diff
@@ -11,9 +11,19 @@ describe("getStatus", () => {
     expect(result).toMatch("success");
   });
 
-  it("returns error if not loading or success and with errorMessage", () => {
-    const result = getStatus({ progress: 100, errorMessage: "An error" });
-    expect(result).toMatch("error");
+  describe("returns error", () => {
```
RohitRajendran, line 14: Both of these tests seem to do the same thing. We can probably revert this to what it was.

packages/components/src/components/FileUploader/FileUploader.stories.tsx
```diff
@@ -209,3 +210,93 @@ export const WithUppy = (): React.ReactElement => {
...
+        type="submit"
+        variant="primary"
+      >
+        Submit
+      </Button>
+    </form>
+  );
+};
+
+export const WithRequiredButton = (): React.ReactElement => {
```
RohitRajendran, line 258: To make it easier for others to understand, can you add a story description for this that explains that if you need to make the file uploader required then you add a required prop to the button component?

Example of how to add a story description:
```
AsControlledRadioGroup.parameters = {
  docs: {
    description: {
      story:
        "If you want to use the RadioGroup with [React Hook Form](https://react-hook-form.com/), use the ControlledRadioGroup component instead.",
    },
  },
};

RohitRajendran (commented):
It's looking good! Just a couple more suggestions to polish it up.

RohitRajendran (approved):
🚀

PR #809, opened 2023-05-10
feat: add withDivider prop to TabsList

rssilva:
## Description of the change
- Add a `withDivider` prop to TabsList to be able to create a element on the bottom of it
- Add a negative zIndex token

Screen Shot 2023-05-10 at 16 44 43

## Type of change

### Development

### Code review


packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -24,6 +24,14 @@ export const Default = (): JSX.Element => {
         <Tab>Tab One</Tab>
         <Tab>Tab Two</Tab>
         <Tab>Tab Three</Tab>
+        {/* <Box.div
```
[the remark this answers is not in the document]
RohitRajendran: I prefer the prop. It feels weird to be passing in a divider element along with the tab elements.

packages/components/src/components/Tabs/TabList.tsx
```diff
@@ -10,21 +10,35 @@ export interface TabListProps extends Omit<TabListPrimitiveProps, "state"> {
...
+
+        {withDivider && (
+          <Box.div
+            backgroundColor="colorBorderWeaker"
+            bottom="0"
+            h="1px"
+            position="absolute"
+            w="100%"
+            zIndex="zIndexNegative1"
+          ></Box.div>
```
RohitRajendran, line 40: Nit:
```suggestion
          />
```

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -228,3 +236,38 @@ export const Fitted = (): JSX.Element => {
     </Tabs>
   );
 };
+
+export const WithDivider = (): JSX.Element => {
```
RohitRajendran, line 240: Typically when creating stories or really doing anything with storybook/component libraries, I like to think about how can I make it easy for someone looking at this to be able to quickly understand and get going with this.

In this case, we could make it easier to understand how exactly to use it with the divider by adding a description to the story that says something like "You can display a divider between the tabs and the content by using the `withDivider` property on the `TabList` component."

(I noticed the docs page isn't working for this component so this won't actually be visible until it's fixed but still good to have for when we do fix it)

.changeset/poor-horses-talk.md
```diff
@@ -0,0 +1,7 @@
+---
+"@localyze-pluto/design-tokens": minor
+"@localyze-pluto/components": minor
+"@localyze-pluto/theme": minor
+---
+
+Add a property to create a divider element on the bottom of tabs list
```
RohitRajendran, line 7: Also should add that the negative 1 token was added for z-index.

RohitRajendran (approved):
Lgtm! Just a couple minor suggestions

PR #816, opened 2023-05-16
feat(components): add onClick to file uploader name

jvaladas:
## Description of the change

Adds compatibility for ~~onClick~~ onClickLabel handlers in the file name anchor of uploaded files. Due to a change in how we are displaying files that don't yet have a public url we needed this handler to change how we're displaying these files when clicking on the anchor link.

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/FileUploader/FileUploader.tsx
+++ b/packages/components/src/components/FileUploader/FileUploader.tsx
@@ -13,7 +13,7 @@ import { FileUploaderProgressBar } from "./FileUploaderProgressBar";
 import { FileUploaderIcon } from "./FileUploaderIcon";
 
 export interface FileUploaderProps
-  extends React.HTMLAttributes<HTMLDivElement> {
+  extends Omit<React.HTMLAttributes<HTMLDivElement>, "onClick"> {
   /** The title of the file */
   label: string;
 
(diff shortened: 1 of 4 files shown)


packages/components/src/components/FileUploader/FileUploader.tsx
```diff
@@ -46,6 +46,9 @@ export interface FileUploaderProps
 
   /** Disables the file uploader */
   disabled?: boolean;
+
+  /** Handles clicking on file name anchor/link */
+  onClick?: (event: React.MouseEvent<HTMLAnchorElement>) => void;
```
nyan07, line 51: Do you think it is a good name for this property? I mean, onClick is pretty generic, and this component has a lot of clickable elements. I wonder if it could create misunderstandings.

apvale, line 51: Can you add a test for this?

jvaladas: I updated for `onLabelClick`, the reason is because the File title may also be a regular `span` element and that way it's more predictable about what this prop does.

jvaladas: Just added, thank you

jvaladas:
No longer needed, I found a solution that doesn't require us to change the FileUploader component in pluto.

PR #825, opened 2023-05-18
feat: add Lucide react icons

rssilva:
## Description of the change
- Add Lucide icons as dependency and a small subset of icons from it

Screen Shot 2023-05-18 at 13 19 51

## Type of change

### Development

### Code review


packages/components/src/components/Icon/Icon.tsx
```diff
@@ -42,8 +44,16 @@ const Icon = React.forwardRef<HTMLDivElement, IconProps>(
...
+    let RenderedIcon = undefined;
+
+    if (isHeroIcon(icon)) {
+      // eslint-disable-next-line import/namespace
+      RenderedIcon = HeroOutlineIcons[icon];
+    }
+
+    if (isLucideIcon(icon)) {
+      RenderedIcon = LucideIcons[icon];
+    }
```
RohitRajendran, line 56: No need to check the second `if` if the first `if` is satisfied.
```suggestion
    if (isHeroIcon(icon)) {
      // eslint-disable-next-line import/namespace
      RenderedIcon = HeroOutlineIcons[icon];
    } else if (isLucideIcon(icon)) {
      RenderedIcon = LucideIcons[icon];
    }
```

packages/components/src/components/Icon/Icon.tsx
```diff
@@ -60,7 +70,9 @@ const Icon = React.forwardRef<HTMLDivElement, IconProps>(
           aria-label={title}
           as={RenderedIcon}
           display="block"
+          h={size}
           verticalAlign="middle"
+          w={size}
```
RohitRajendran, line 75: Just for my understanding, why was this change necessary?

packages/components/src/components/Icon/isHeroIcon.ts
```diff
@@ -0,0 +1,8 @@
+import * as HeroOutlineIcons from "@heroicons/react/24/outline";
+import { IconName } from "./IconName";
+
+export const isHeroIcon = (
+  name: IconName
+): name is keyof typeof HeroOutlineIcons => {
```
RohitRajendran, line 6: Isn't this just returning a boolean? Could you help me understand `name is keyof typeof HeroOutlineIcons`?

packages/components/src/components/Icon/HeroIconName.ts
RohitRajendran, line 1: What do you think about moving the new files into subfolders like `/types` and `/utils` or something to make the file structure a little cleaner and easier to parse? It would make it easier to find and pick out Icon.tsx as the main component file when looking at the folder.

.changeset/popular-plums-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@localyze-pluto/components": minor
+---
+
+Add a Lucide Icons as dependency with a specific subset available
```
RohitRajendran, line 5: We could be a little more descriptive in this changeset and list the new icon options available.

packages/components/src/components/Icon/IconName.ts
```diff
@@ -0,0 +1,4 @@
+import { HeroIconName } from "./HeroIconName";
+import { LucideIconName } from "./LucideIconName";
+
+export type IconName = HeroIconName | LucideIconName;
```
RohitRajendran, line 4: There's a few other components that currently use heroicons: Badge, Button, Input. Shouldn't they be updated to have all the same icon options?

packages/components/src/components/Icon/Icon.tsx
rssilva: if we don't apply those, lucide icons will have a default value of 24 fitting the container

Screen Shot 2023-05-19 at 15 29 12

packages/components/src/components/Icon/isHeroIcon.ts
rssilva: I believe this is a safe(r) method to check if an item is present as key of an object? https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

packages/components/src/components/Icon/IconName.ts
rssilva: oh! Good call!

packages/components/package.json
```diff
@@ -51,6 +51,7 @@
     "ariakit": "^2.0.0-next.43",
     "html-react-parser": "^3.0.9",
     "lodash": "^4.17.21",
+    "lucide-react": "^0.218.0",
```
RohitRajendran, line 54: Should this be in both `dependencies` and `peerDependencies`? At least based on how `@heroicons/react`, not sure if that's actually needed.

.changeset/popular-plums-deny.md
```diff
@@ -2,4 +2,4 @@
 "@localyze-pluto/components": minor
 ---
 
-Add a Lucide Icons as dependency with a specific subset available
+Add a Lucide Icons as dependency with a specific subset available: baggage-claim, fingerprint, heart-handshake, palette, slack
```
RohitRajendran, line 5: Won't we now have `fingerprint` from both Lucide and HeroIcons? Shouldn't we consolidate on 1 so that it's not inconsistently used?

rssilva: good call!

packages/components/package.json
RohitRajendran: Ah, my bad!

PR #877, opened 2023-06-16
Fix turbo lint

RohitRajendran:
## Description of the change

1. Updates packages to fix the `TypeError: sourceCode.getScope is not a function` we were getting when running `yarn lint` with `turbo lint`. 
2. Added support for running `yarn prettier-check` with `turbo`
3. Fix eslint errors (Thanks @rssilva!)
4. Updated the workflow to use `yarn lint` and `yarn prettier-check` so that we no longer have the discrepancies between linter errors showing up differently depending on workflow or local

## Testing the change

- [ ] Workflow should pass
- [ ] Can run `yarn lint` and `yarn prettier-check` locally

## Type of change

### DIFF
--- a/.changeset/chilly-spiders-cry.md
+++ b/.changeset/chilly-spiders-cry.md
@@ -1,0 +1,5 @@
+---
+"@localyze-pluto/components": patch
+---
+
+Fix eslint configuration, fix lint errors on Dropzone and ModalBody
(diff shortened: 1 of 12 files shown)


rssilva (approved):
That's way better and it's working on VSCode now! 🚢

PR #988, opened 2023-08-31
[OCTO-576] implement design token storybook panel in pluto ds

blumaa:
## Description of the change

Here we are adding a preview of all of our design system tokens

Screenshot 2023-09-01 at 17 15 24
Screenshot 2023-09-01 at 17 14 51
Screenshot 2023-09-01 at 17 15 31
Screenshot 2023-09-01 at 17 15 37
Screenshot 2023-09-01 at 17 15 49
Screenshot 2023-09-01 at 17 15 58

## Testing the change

## Type of change

### Development

### Code review


.changeset/tiny-seals-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@localyze-pluto/design-tokens": minor
+---
+
```
nyan07, line 4: Should we have it is only docs?

packages/components/src/design-token-preview-tables/ColorTokensTable.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { render, screen } from "@testing-library/react";
```
nyan07, line 1: Does it make sense to write tests for docs?

nyan07 (commented):
Do we need all those components?
I was expecting a few reused on the `.mdx` files.

.changeset/tiny-seals-guess.md
blumaa: @nyan07 Can you explain more? I wasn't sure about the changeset here. Mostly, the docs components are importing from the components folder. So should the localyze-pluto/components have a minor update here?

packages/components/src/design-token-preview-tables/ColorTokensTable.test.tsx
blumaa: I think so? At least for these ones since they are being 'rendered' as jsx components.

.changeset/tiny-seals-guess.md
nyan07: We don't need to generate a new version because it doesn't change anything on components that are exported to usage. I believe the idea here is to only focus on docs and not change the bundle. So, it is just a change on the repo, not a new version.

packages/components/package.json
```diff
@@ -34,6 +34,7 @@
     "@radix-ui/react-radio-group": "^1.1.1",
     "@radix-ui/react-switch": "^1.0.1",
     "@radix-ui/react-toast": "^1.1.2",
+    "@storybook/addon-docs": "^7.3.2",
```
nyan07, line 37: Do we need this change on the components package?

packages/design-tokens/design-token-previews/utils.tsx
```diff
@@ -0,0 +1,117 @@
+import React from "react";
```
nyan07, line 1: As we have a utils folder, shouldn't it be inside it withe a more descriptive name?

packages/design-tokens/design-token-previews/utils.tsx
```diff
@@ -0,0 +1,117 @@
...
+    <div style={{ display: "flex" }}>
+      <div>{tokenName}</div>
+      <div>
+        <CopyToClipboardButton textToCopy={tokenName} />
+      </div>
+    </div>
+  );
+};
+
+export const TOKENS = {
```
nyan07, line 58: Shouldn't we keep it in a constants file?

nyan07: I also think we should rethink this variable name. It isn't about tokens but columns or transformations.

packages/design-tokens/design-token-previews/utils/previews.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+import includes from "lodash/includes";
+
+import { TokenEntry } from "../types/Token";
+
+import { FontSizePreview } from "../FontSizePreview";
+import { FontWeightPreview } from "../FontWeightPreview";
+import { IconSizePreview } from "../IconSizePreview";
+
+export const createFontWeightPreview = ([
```
nyan07, line 11: Do we need this file, or would it be better to create anonymous functions on the `constants.tsx`. e.g., 

```typescript
  [getKey(fontWeightTokens)]: [
    { name: "Pixels", transform: getValue },
    { name: "Preview", transform: ([tokenSuffix, token]: TokenEntry) => <FontWeightPreview fontWeight={token.value} key={tokenSuffix } />},
  ],
```

packages/design-tokens/design-token-previews/utils.tsx
```diff
@@ -0,0 +1,117 @@
...
+    },
+    {
+      name: "Hsla",
+      transform: ([, token]: TokenEntry): string => hexToHsla(token.value),
+    },
+    { name: "Preview", transform: createColorPreview },
+  ],
+};
+
+export const renderRows = (
```
nyan07, line 104: I would suggest giving this function a different name because it isn't rendering but parsing the token data. The render happens on the `TokensTable` component (after iterating on this function result).

packages/design-tokens/design-token-previews/utils/previews.tsx
blumaa: To me it makes more sense to keep this file. If we put them all in the constants.ts it feels to messy to me.

packages/design-tokens/design-token-previews/utils/previews.tsx
```diff
@@ -0,0 +1,94 @@
...
+  tokenSuffix,
+  token,
+]: TokenEntry): JSX.Element => {
+  return <FontWeightPreview fontWeight={token.value} key={tokenSuffix} />;
```
nyan07, line 15: Why do we need a component for this one and not use it directly as you did with the others?

packages/design-tokens/design-token-previews/utils/previews.tsx
```diff
@@ -0,0 +1,94 @@
...
+        backgroundColor: token.value,
+      }}
+    >
+      &nbsp;
+    </div>
+  );
+};
+
+export const createSpacePreview = ([, token]: TokenEntry): JSX.Element => {
+  if (includes(token.value, "-")) {
```
nyan07, line 73: Why do we need it?

packages/design-tokens/design-token-previews/utils/previews.tsx
```diff
@@ -0,0 +1,94 @@
...
+import includes from "lodash/includes";
+
+import { TokenEntry } from "../types/TokenEntry";
+
+import { FontSizePreview } from "../FontSizePreview";
+import { FontWeightPreview } from "../FontWeightPreview";
+import { IconSizePreview } from "../IconSizePreview";
+
+export const createFontWeightPreview = ([
+  tokenSuffix,
```
nyan07, line 12: Does it make sense to create components for things you are not reusing?
If you want to have review components I would create one per type. e.g., reusing font size and weight as they render the same content (we change just the prop).
Same for borders. We are just changing the prop.


Rethinking it. I think we need only;y for icons. The rest could be all the same.

packages/design-tokens/design-token-previews/constants.ts
```diff
@@ -0,0 +1,70 @@
...
+  createFontSizePreview,
+  createIconSizePreview,
+  createFontWeightPreview,
+  createBorderStylePreview,
+  createBorderWidthPreview,
+} from "./utils/previews";
+
+import { hexToHsla, hexToRgb } from "./utils/colors";
+import { getComment, getKey, getValue } from "./utils/utils";
+
```
nyan07, line 25:
```suggestion
const buildBox = (props) => <Box.div {...props} />;
const buildText = (props) => <Text.span {...props} />;
```

packages/design-tokens/design-token-previews/constants.ts
```diff
@@ -0,0 +1,70 @@
...
+export const TOKEN_COLUMNS = {
+  [getKey(borderRadiiTokens)]: [
+    { name: "Pixels", transform: getValue },
+    { name: "Preview", transform: createBorderRadiiPreview },
```
nyan07, line 29:
```typescript
const createBox = (tokenAttribute, defaultProps = {}) => ([tokenSuffix, token] => {
  const props = {
    ...defaultProps,
    [tokenAttrribute]: token.value
  };
  return <Box.div {...props}>
}
```


```suggestion
    { name: "Preview", transform: createBox('borderRaius') },
```

nyan07: Same for all the other items. You may need to create a `createText` if the `BOx.div` doesn't accept font properties. I can't remember.

packages/design-tokens/design-token-previews/utils/previews.tsx
nyan07: https://github.com/Localitos/pluto/pull/988/files#r1321781393

packages/design-tokens/design-token-previews/utils/utils.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const getValue = ([, token]: TokenEntry): string => {
+  return token.value;
+};
+
+export const getComment = ([, token]: TokenEntry): string => {
+  return token.comment || "";
+};
+
+export const formatTokenName = (
```
nyan07, line 22: Shouldn't it be a `TokenName` component?

packages/design-tokens/design-token-previews/utils/utils.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
```
nyan07, line 1: I wouldn't use a utils file inside a utils folder. COnsider my other comment on this file to see if you need to address this one.

packages/design-tokens/design-token-previews/constants.ts
nyan07: With this approach, we would only need some of the preview components. And I would consider putting them all inside a folder to nox mix with the `.mdx`

nyan07 (approved):
Please look into this comment https://github.com/Localitos/pluto/pull/988/files#r1321781393 before all the others. I think a lot of files can be removed following this approach.

packages/design-tokens/design-token-previews/types/TokenTable.ts
```diff
@@ -0,0 +1,8 @@
+import { TokenEntry } from "./TokenEntry";
```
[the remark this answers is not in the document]
blumaa: @nyan07 Not sure what you mean here. Are you talking about the name of the types?

packages/design-tokens/design-token-previews/utils/previews.tsx
blumaa: @nyan07 Not exactly sure what you mean here. This conditional only renders the preview when the space token is NOT negative. So, the negative space tokens actually don't preview anything. I am not sure how to preview/display negative tokens.

nyan07: A token can be `-10px`,  `-20px`, etc. How people will know the difference without checking the value?
I belive they should have the same preview as a positive one, but with the minus on the value column.

packages/design-tokens/design-token-previews/types/TokenTable.ts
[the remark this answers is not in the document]
nyan07: The file name is `TokenTable`. What you have inside the file is actually `Row` and `Column` types. I would use different files with the types' names or keep them inside the `TokensTable` component if it isn't reused in other places.

packages/components/src/design-token-preview-tables/ColorTokensTable.test.tsx
nyan07: I'm not sure because they are not used in the app. They are components used only on docs they won't break anything on the apps. IMO it looks like testing the tests or components we build to use on stories.

nyan07:
What do you think about having a single line with the same space on top/bottom instead of a square?

image

packages/design-tokens/design-token-previews/types/TokenColumn.ts
```diff
@@ -0,0 +1,6 @@
+import { TokenEntry } from "./TokenEntry";
+
+export type Column = {
```
nyan07, line 3:
```suggestion
export type TokenColumn = {
```

packages/design-tokens/design-token-previews/types/TokenRow.ts
```diff
@@ -0,0 +1 @@
+export type Row = Array<JSX.Element | string>;
```
nyan07, line 1:
```suggestion
export type TokenRow = Array<JSX.Element | string>;
```

packages/design-tokens/design-token-previews/utils/createFontSizePreview.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
```
nyan07, line 1: Why do we still need this one? Wouldn't it be possible to use the `createBox`?

packages/design-tokens/design-token-previews/utils/constants.tsx
```diff
@@ -0,0 +1,111 @@
...
+  hexToHsla,
+} from "../utils";
+
+const EXAMPLE_TEXT = "The quick brown fox jumped over the lazy dog.";
+
+export const TOKEN_COLUMNS = {
+  [getTokenKey(borderRadiiTokens)]: [
+    {
+      name: "Name",
+      transform: getTokenName(getTokenKey(borderRadiiTokens)),
```
nyan07, line 30: I see two options to not repeat this line so many times.

```typescript
transform: getTokenName(borderRadiiTokens),

// packages/design-tokens/design-token-previews/utils/getTokenName.tsx
export const getTokenName =
  (tokens: Record<string, Record<string, Token>) =>
  // eslint-disable-next-line react/display-name
  ([suffix]: TokenEntry): JSX.Element => {
    const prefix = getTokenKey(tokens);
    const tokenName = camelCase(`${prefix}${capitalize(suffix)}`);
    return <TokenName tokenName={tokenName} />;
  };

```

OR

```typescript
const getTokenName =  (tokens: Record<string, Record<string, Token>) => {
  const prefix = getTokenKey(tokens);
  return getTokenNameComponent(prefix);
}


// packages/design-tokens/design-token-previews/utils/getTokenName.tsx
// just renaming it. I suggested `getTokenNameComponent` on the code above, but I don't think it is the best name.
```

packages/design-tokens/design-token-previews/components/createBox.tsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import includes from "lodash/includes";
+import replace from "lodash/replace";
+import { TokenEntry } from "../types/TokenEntry";
+import { Box } from "../../../components/src/primitives/Box";
+
+export const createBox =
+  (tokenAttribute: string, defaultProps = {}, text = "\0") =>
```
nyan07, line 8: The `text` argument is just an exception for font preview. Shouldn't we use a single front preview? Or a `createFrontPreview` function? I left a similar suggestion in another file.

packages/design-tokens/design-token-previews/components/TokenName.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Box } from "../../../components/src/primitives/Box";
+import { CopyToClipboardButton } from "./CopyToClipboardButton";
+
+export const TokenName = ({
+  tokenName,
+}: {
+  tokenName: string;
+}): JSX.Element => {
+  // const tokenName = camelCase(`${prefix}${capitalize(suffix)}`);
```
nyan07, line 10: Should this comment be removed?

packages/design-tokens/design-token-previews/components/formatTokenName.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
nyan07, line 1: We already have something like this on the `TokenName` component. Are we still using this function?

packages/design-tokens/design-token-previews/components/IconSizePreview.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { SystemProp, Theme } from "@xstyled/styled-components";
+import { Icon } from "../../../components/src/components/Icon/Icon";
+import { ThemeProvider, theme } from "../../../theme";
+
+const IconSizePreview = ({ pixels }: { pixels: string }): JSX.Element => {
```
nyan07, line 6:
```suggestion
const IconPreview = (props: IconProps): JSX.Element => {
```

packages/design-tokens/design-token-previews/components/IconSizePreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <ThemeProvider theme={theme}>
+      <Icon
```
nyan07, line 9:
```suggestion
      <Icon
        {...props}
        // ...
```

packages/design-tokens/design-token-previews/components/IconSizePreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+        decorative
+        display="flex"
+        icon="HeartIcon"
+        size={pixels as SystemProp<keyof Theme["sizes"], Theme>}
```
nyan07, line 13:
```suggestion
```

packages/design-tokens/design-token-previews/components/IconSizePreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+      />
+    </ThemeProvider>
+  );
+};
```
nyan07, line 17: Following this suggestion, you don't need to have it specifically for size and you could also set the other props externally. The only reason to have this component IMO is only the `ThemeProvider`. 

```suggestion
const IconPreview = (props: IconProps): JSX.Element => {
  return (
    <ThemeProvider theme={theme}>
      <Icon
        {...props}
        decorative
        display="flex"
        icon="HeartIcon"
      />
    </ThemeProvider>
  );
};
```

Another thing to consider is, as we already have the `createIconSizePreview` function, remove this component and return the `ThemeProvider` + `Icon` there. Which seems better to me because the function is small and we are not reusing this component anywhere.

packages/design-tokens/design-token-previews/borders.mdx
```diff
@@ -0,0 +1,22 @@
+import borderStyleTokens from "../src/tokens/border-style.tokens.json";
```
nyan07, line 1: Not specifically for this file. It is more related to the folder. We are already inside the `design-tokens` folder. Calling this folder `design-tokens-previews` it is redundant. 

Following what we have on [storybook](https://storybook.js.org/docs/react/writing-docs/mdx) for mdx use case, I would call it `docs`.

packages/design-tokens/design-token-previews/components/TokensTable.tsx
```diff
@@ -0,0 +1,58 @@
...
+              {map(row, (cell) => {
+                return (
+                  <Td key={isString(cell) ? cell : `${cell.props.tokenName}`}>
+                    {cell}
+                  </Td>
+                );
+              })}
+            </Tr>
+          );
+        })}
```
nyan07, line 54: It generates keys with `undefined`.

image

Relying on `row[0].props.tokenName` seems complicated because it can be any react component and not have the token name.

It may be better to keep only the string on the first column, use it as a key, and put the component here instead. 

```suggestion
        {map(rows, (row) => {
          const tokenName = row[0];
          return (
            <Tr key={tokenName}>
              {map(row, (cell, index) => {
              const column = columns[index];
                return (
                  <Td key={`${tokenName}${column}`}>
                    {index === 0 ? <TokenName name={cell} /> : cell}
                  </Td>
                );
              })}
            </Tr>
          );
        })}
```

I would also change the `Row` type to guarantee that only the last item is a react element.

```typescript
type TokenRow = [...string[], JSX.ELement];
```

packages/design-tokens/design-token-previews/utils/constants.tsx
```diff
@@ -0,0 +1,111 @@
+import fontSizeTokens from "../../src/tokens/font-size.tokens.json";
```
nyan07, line 1: It isn't a utils file but the constant that defines the data for the entire table, I would keep it on the `TokenTable` level.

packages/design-tokens/design-token-previews/utils/constants.tsx
nyan07, line 30: I would consider having a square passing the optional props `{ w: "50px", h: "50px"}`, for example. Because we have a circle. 

We would benefit from some margin too as the items seem too close. I left a comment with the screenshot before, but I believe you missed it.

blumaa:
> What do you think about having a higher box with more spacing on the top and bottom previews? They seem too close.
> 
> image

@nyan07 What does it mean a 'higher box'? What seems too close?

packages/design-tokens/design-token-previews/utils/createFontSizePreview.tsx
blumaa: It's not possible because we have to use the fontSize tokens from the theme in the `Box.div`

packages/design-tokens/design-token-previews/utils/getTokenKey.ts
```diff
@@ -0,0 +1,5 @@
+import keys from "lodash/keys";
+
+export const getTokenKey = (object: Record<string, unknown>): string => {
```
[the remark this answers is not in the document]
blumaa: I don't think this is the same. At least it throws a lot of errors when I change it to Token.

sliminas (approved):
Looks great! Thanks a lot! :tada:

packages/design-tokens/design-token-previews/components/IconSizePreview.tsx
blumaa: I tried to implement this with Stephan and we couldn't figure out what to do here. In this case, it's working so I wouldn't change it.

packages/design-tokens/design-token-previews/components/TokensTable.tsx
blumaa: This one didn't make sense to me how to implement. Because sometimes the key of a color Token could be duplicate and not unique

packages/design-tokens/design-token-previews/utils/constants.tsx
blumaa: I'm not sure about this one. I made them with the height and width but we also have a pill shape and that should not be a square.

nyan07 (approved):
Amazing work! :rocket:

PR #1026, opened 2023-10-05
feat: add menu component

sliminas:
## Description of the change

Add a menu component (dropdown).

---

image

---

https://github.com/Localitos/pluto/assets/6059188/497b3950-3bb4-4dbb-abf5-a7533b523e6a

## Type of change

### DIFF
--- a/packages/components/src/components/Menu/Menu.stories.tsx
+++ b/packages/components/src/components/Menu/Menu.stories.tsx
@@ -1,16 +1,20 @@
 import type { Meta } from "@storybook/react";
+import { StoryObj } from "@storybook/react";
 import React from "react";
 import noop from "lodash/noop";
 import { Button } from "../Button";
 import { Box } from "../../primitives/Box";
 import { Menu, MenuItemProps } from "./Menu";
 
-export default {
-  component: Menu,
+const meta: Meta<typeof Menu> = {
   title: "Components/Menu",
-} as Meta<typeof Menu>;
+  component: Menu,
+};
 
-export const Default = (): JSX.Element => {
+export default meta;
+type Story = StoryObj<typeof Menu>;
+
+const DefaultMenu = (): JSX.Element => {
   const items: MenuItemProps[] = [
     {
       onClick: noop,
@@ -29,7 +33,18 @@ export const Default = (): JSX.Element => {
   );
 };
 
-export const WithCustomMenuButton = (): JSX.Element => {
+export const Default: Story = {
+  render: (): JSX.Element => <DefaultMenu />,
+};
+
+Default.parameters = {
+  docs: {
+    storyDescription:
+      "A menu is a button element that opens a menu with items.",
+  },
+};
+
+const CustomButtonMenu = (): JSX.Element => {
   const items: MenuItemProps[] = [
     {
       onClick: noop,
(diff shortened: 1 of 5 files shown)


packages/components/src/components/Menu/Menu.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { Button } from "../Button";
+import { Box } from "../../primitives/Box";
+import { Menu, MenuItemProps } from "./Menu";
+
+export default {
+  component: Menu,
+  title: "Components/Menu",
+} as Meta<typeof Menu>;
+
+export const Default = (): JSX.Element => {
```
blumaa, line 13: This is an old way of writing stories using CSF2. Can we use the CSF3 methods here please? https://storybook.js.org/docs/react/writing-stories/introduction

Or you could look at packages/components/src/components/Pagination/Pagination.stories.tsx for examples

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export type MenuProps = {
+  /** The button element to open the menu and display the list of menu items. */
+  menuButton?: JSX.Element;
+  /** The list of menu items. */
+  items: MenuItemProps[];
+};
+
+const EllipsisButton = (
```
blumaa, line 25: I wonder if we could call this "VerticalEllipsisButton"

In mobile, web, and general application design, the vertical ellipsis, ⋮, is sometimes used as an interface element, where it is sometimes called a [kebab icon](https://en.wikipedia.org/wiki/Kebab_icon). [source](https://en.wikipedia.org/wiki/Ellipsis#:~:text=In%20mobile%2C%20web%2C%20and%20general,sometimes%20called%20a%20kebab%20icon.)

blumaa: Or "KebabButton" :)

blumaa:
It looks great Stephan! Well done.

Can we add a story with more than two (at least five options) to see how that looks?

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,84 @@
...
+          {(buttonProps: DisclosureProps) =>
+            React.cloneElement(button, buttonProps)
+          }
+        </MenuButton>
+
+        <AriakitMenu state={state}>
+          <Box.div
+            borderColor="colorBorderPrimary"
+            borderRadius="borderRadius20"
+            borderStyle="borderSolid"
```
blumaa, line 54:
```suggestion
            borderStyle="borderStyleSolid"
```

sliminas: I tried that before, but it's an invalid value:
image

blumaa: That is strange because for me I have the opposite problem. Maybe @rssilva or @apvale could see what they get?

blumaa (approved):
Except for the borderStyleSolid problem, this looks good to me.

packages/components/src/components/Menu/Menu.tsx
sliminas: `yarn build` resolved the issue.

PR #1058, opened 2023-10-23
[components] fix: import uppy differently, add trigger to dependency arrays

rssilva:
## Description of the change
- Fix lint warnings 

## Type of change

### Development

### DIFF
--- a/packages/components/src/components/FileUploader/FileUploader.stories.tsx
+++ b/packages/components/src/components/FileUploader/FileUploader.stories.tsx
@@ -1,6 +1,6 @@
 import type { ComponentMeta, ComponentStory } from "@storybook/react";
 import React, { useState } from "react";
-import UppyCore from "@uppy/core";
+import { Uppy } from "@uppy/core";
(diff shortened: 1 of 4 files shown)


packages/components/src/components/FileUploader/FileUploader.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import type { ComponentMeta, ComponentStory } from "@storybook/react";
 import React, { useState } from "react";
-import Uppy from "@uppy/core";
+import UppyCore from "@uppy/core";
```
padulafacundo, line 3: Would using the explicit import fix it?
```suggestion
import { Uppy } from "@uppy/core";
```

padulafacundo (approved):

packages/components/src/components/FileUploader/FileUploader.stories.tsx
rssilva: way better! Good call!

PR #1086, opened 2023-10-25
feat(components): add disabled props to Menu MenuItems

blumaa:
## Description of the change

We need to be able to disable the buttons in the menu. Therefore I've added the disabled prop to the menu items here.

## Testing the change
Tests were added for this

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Menu/Menu.test.tsx
+++ b/packages/components/src/components/Menu/Menu.test.tsx
@@ -36,4 +36,21 @@ describe("<Menu />", () => {
     expect(screen.getByText("Item 1")).toBeInTheDocument();
     expect(screen.getByText("Item 2")).toBeInTheDocument();
   });
+
+  it("allows a disabled prop for a menu item and disables the button", () => {
+    render(
+      <Menu items={[{ onClick: jest.fn(), label: "Item 1", disabled: true }]} />
+    );
+
+    const menuButton = screen.getByRole("button");
+    expect(menuButton).toBeInTheDocument();
+
+    userEvent.click(menuButton);
+
+    expect(screen.getByRole("presentation")).toBeInTheDocument();
+
+    expect(
+      screen.getByRole("button", { name: /item 1/i, hidden: true })
+    ).toBeDisabled();
+  });
 });
--- a/packages/components/src/components/Menu/Menu.test.tsx
+++ b/packages/components/src/components/Menu/Menu.test.tsx
@@ -38,9 +38,8 @@ describe("<Menu />", () => {
   });
 
   it("allows a disabled prop for a menu item and disables the button", async () => {
-    render(
-      <Menu items={[{ onClick: jest.fn(), label: "Item 1", disabled: true }]} />
-    );
+    const onClick = jest.fn();
+    render(<Menu items={[{ onClick, label: "Item 1", disabled: true }]} />);
 
     const menuButton = screen.getByRole("button");
     expect(menuButton).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Menu/Menu.test.tsx
```diff
@@ -36,4 +36,21 @@ describe("<Menu />", () => {
...
+
+  it("allows a disabled prop for a menu item and disables the button", () => {
+    render(
+      <Menu items={[{ onClick: jest.fn(), label: "Item 1", disabled: true }]} />
+    );
+
+    const menuButton = screen.getByRole("button");
+    expect(menuButton).toBeInTheDocument();
+
+    userEvent.click(menuButton);
```
rssilva, line 48: I believe that we need an await here, right?

packages/components/src/components/Menu/Menu.test.tsx
```diff
@@ -36,4 +36,21 @@ describe("<Menu />", () => {
     expect(screen.getByText("Item 1")).toBeInTheDocument();
     expect(screen.getByText("Item 2")).toBeInTheDocument();
   });
+
+  it("allows a disabled prop for a menu item and disables the button", () => {
+    render(
+      <Menu items={[{ onClick: jest.fn(), label: "Item 1", disabled: true }]} />
```
rssilva, line 42: Small ignorable suggestion here. You could:

```
const onClick = jest.fn()

<Menu items={[{ onClick, label: "Item 1", disabled: true }]} />
```

then you can be sure that `onClick` is not called

rssilva (approved):
looks good! Just a check on the necessity of that `await` before merging

packages/components/src/components/Menu/Menu.test.tsx
apvale: Yeah, we need it. Do you guys think we should have this as a lint error? 🤔 
I've seen more cases like this.

apvale: Anyway, I can start the discussion in the next Pluto's meeting. 🙏🏻

packages/components/src/components/Menu/Menu.test.tsx
apvale: Agree with you, but I would create another test just to test the onClick prop and make sure it's only called when clicked.

packages/components/src/components/Menu/Menu.test.tsx
```diff
@@ -36,4 +36,21 @@ describe("<Menu />", () => {
...
+
+    const menuButton = screen.getByRole("button");
+    expect(menuButton).toBeInTheDocument();
+
+    userEvent.click(menuButton);
+
+    expect(screen.getByRole("presentation")).toBeInTheDocument();
+
+    expect(
+      screen.getByRole("button", { name: /item 1/i, hidden: true })
```
apvale, line 53: Why do you need the hidden true here?

sliminas (approved):

packages/components/src/components/Menu/Menu.test.tsx
blumaa: I'm not exactly sure @apvale . But the button is hidden in the dom for some reason. Although I can see it in story book and in the UI, when testing, it is hidden.

blumaa: Well I figured out that it is because of the async/await. It works when I removed hidden and I added the async/await with the user click.

PR #1099, opened 2023-11-06
chore: allow any space token value as marginBottom to Heading component

rssilva:
## Description of the change
- Allow more space tokens to be marginBottom on the Heading component

## Type of change

### Development

### DIFF
--- a/.changeset/afraid-cats-marry.md
+++ b/.changeset/afraid-cats-marry.md
@@ -1,0 +1,5 @@
+---
+"@localyze-pluto/components": patch
+---
+
+Allow any space tokens to be used as margin bottom on heading component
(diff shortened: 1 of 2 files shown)


.changeset/afraid-cats-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@localyze-pluto/components": patch
+---
+
+Allow any space tokens to be used as margin bottom on heading component
```
apvale, line 5: I think the change would be not to allow any margin to this component. So, by saying that and to avoid any issue with the code that already exists, we can set the marginBottom as space0 by default. I'm saying that because we are always setting marginBottom as "space0" because the component is setting margin when it shouldn't.

PR #1107, opened 2023-11-08
Re-use Modal for Alert component

sliminas:
## Description of the change

## Testing the change

## Type of change

### Development

### Code review


packages/components/src/components/Modal/ModalHeader.tsx
```diff
@@ -14,6 +14,7 @@ const ModalHeader = React.forwardRef<HTMLDivElement, ModalHeaderProps>(
   ({ children, ...props }, ref) => {
     return (
       <Box.div
+        alignItems="center"
```
sliminas, line 17: The heading and the close button were not on one line before.

.changeset/khaki-brooms-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"@localyze-pluto/components": major
+---
+
+Remove the `AlertDialog` component and replace it with the `ConfirmationModal` component which re-uses the `Modal` component to unify the appearance of modals
```
nyan07, line 5: Are we using the `AlertDialog`  somewhere? Shouldn't we deprecate the component first and them remove it in the next version?

packages/components/src/components/ConfirmationModal/ConfirmationModal.test.tsx
```diff
@@ -0,0 +1,88 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
```
nyan07, line 2: Could you please fix the linter here?

packages/components/src/components/ConfirmationModal/ConfirmationModal.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+const CustomModal = () => {
+  const state = useModalState();
+
+  return (
+    <div>
+      <button onClick={state.show}>{OPEN_MODAL_TEXT}</button>
+
+      <ConfirmationModal
+        buttonLabel="Delete"
+        buttonVariant="destructive"
```
nyan07, line 34: Shouldn't it be a ConfirmationDialog variant instead of just a button variant? Do we have a definition from design side?

packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const ConfirmationModal = React.forwardRef<
+  HTMLDivElement,
+  ConfirmationModalProps
+>(
+  (
+    {
+      state,
+      buttonVariant = "primary",
+      buttonLabel = "Confirm",
```
nyan07, line 33: For the future: Do we want to keep default text on Pluto components? Or, how should we manage them?
We are talking about supporting multiple languages since it is a requisite for blue-collar companies, so it can become a topic soon.

packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+const ConfirmationModal = React.forwardRef<
+  HTMLDivElement,
+  ConfirmationModalProps
+>(
+  (
+    {
+      state,
+      buttonVariant = "primary",
```
nyan07, line 32: Inspired by [atomic design](https://bradfrost.com/blog/post/atomic-web-design/#molecules), a beneficial framework for design systems, I prefer to have a version customized for the component, considering that multiple aspects of it may undergo changes.

Currently, the focus is on the button, but it's important to note that this component is part of a larger structure that should change cohesively. We should avoid controlling specific components through props and, instead, ensure that the output maintains consistency for combinations that require it.

packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
nyan07: Based on my previous comment https://github.com/Localitos/pluto/pull/1107/files#r1392326834 I would also suggest renaming it to `confirmationLabel` or `confirmationText`.

packages/components/src/components/ConfirmationModal/ConfirmationModal.test.tsx
sliminas: I would rather like to do it in the whole repo in a separate PR.

.changeset/khaki-brooms-clean.md
sliminas: We have a major version bump so when we upgrade to the next version we can replace the `AlertDialog` with the `ConfirmationModal`.

packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
sliminas: > For the future: Do we want to keep default text on Pluto components? Or, how should we manage them?

I don't see a problem with that. If we have a default state of a component it should also have a default label.
When we introduce I18n then we need to adjust for that when we need this feature..

packages/components/src/components/ConfirmationModal/ConfirmationModal.test.tsx
sliminas: See https://github.com/Localitos/pluto/pull/1112

nyan07 (approved):

PR #1126, opened 2023-11-24
fix: avoid running commitlint until it is fixed

rssilva:
## Description of the change
- Avoid running commitlint until it's fixed

## Type of change


apvale:
@rssilva Why do we need a changeset for this? Shouldn't this package only affect Pluto internally?

rssilva:
@apvale I want to run a release on purpose and check if the fix works

apvale (approved):

PR #1156, opened 2023-12-05
fix: reverts storybook avoiding broken build

rssilva:
## Description of the change

## Type of change


apvale (approved):
Do we need a changeset for this? 🤔

PR #1184, opened 2023-12-15
chore: add optional callback to Confirmation Modal cancel button

rssilva:
## Description of the change
- Add onCancel optional callback to Confirmation Modal 

## Type of change

### Development

### DIFF
--- a/packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
@@ -9,6 +9,7 @@ export interface ConfirmationModalProps {
   confirmLabel?: string;
   destructive?: boolean;
   onConfirm: () => void;
+  onCancel?: () => void;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/components/src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -9,6 +9,7 @@ export interface ConfirmationModalProps {
   confirmLabel?: string;
   destructive?: boolean;
   onConfirm: () => void;
+  onCancel?: () => void;
```
apvale, line 12: Why do you think we need a callback for this?

rssilva: To trigger actions (like track events, for example) when the user clicks on the cancel button

apvale: I meant why changing this in the PR? Is there any specific scenario that would be necessary to override the state.hide action?

rssilva: The state.hide() will still be called here

apvale (approved):

PR #1207, opened 2023-12-27
chore: Add Calendar icon from lucide package

rssilva:
## Description of the change

- Add calendar icon from the Lucide

## Testing the change

- [ ] Write your testing instructions here

## Type of change

- [x] Non-Breaking Change (change to existing functionality)

PR #1206, opened 2023-12-27
feat: Add Noto Sans font

rssilva:
## Description of the change
Add Noto Sans font to the Design System

## Type of change

- [x] New feature (non-breaking change that adds functionality)

### Development

- [x] The code changed/added as part of this pull request has been covered with tests
- [x] All tests related to the changed code pass in development

PR #1209, opened 2023-12-28
feat(components): add Content Card component

rssilva:
## Description of the change
- Add Content Card component
- This PR depends on #1206, #1207 and #1208 

Screen Shot 2024-01-05 at 09 45 37

## Type of change

### Development

### Code review


packages/components/src/components/ContentCard/ContentCard.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+        href=""
+        imageAlt=""
+        imageSrc=""
+        tag="🏠 housing"
+        text="hello! Long text"
+        title="Good title!"
+      />,
+    );
+
+    expect(screen.queryByText("Good title!")).toBeVisible();
```
apvale, line 22: Why using "queryByText"? Why "getByText" is not enough? 🤔

apvale (commented):
It looks good already. 

I have a suggestion: 

What do you think of having the Image as a component and making it more composible? For example: 

```
<ContentCard position="left">
   <Image alt={} src={} />
<ContentCard />
```

In this way, we allow the Developer to select the image position and we remove some props from the parent component. 

What do you think?

packages/components/src/components/ContentCard/ContentCard.test.tsx
rssilva: oh duh me

packages/components/src/components/ContentCard/ContentCard.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+        href=""
+        imageAlt="a test image"
+        imageSrc="path-to-image.jpg"
+        tag=""
+        text=""
+        title=""
+      />,
+    );
+
+    expect(screen.getByAltText("a test image")).toHaveAttribute(
```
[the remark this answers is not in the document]
rssilva: This doesn't work because the link does not have a src attribute. That belongs to the image

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+  imageSrc,
+  title,
+  tag,
+  text,
+  date,
+  buttonText,
+  onClickButton,
+  linkText,
+  linkHref,
+  imageAlt,
```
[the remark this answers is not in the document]
rssilva: I was thinking that this was only changing according the screen size. Working on it

packages/components/src/components/ContentCard/ContentCard.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <ContentCard
+        date=""
+        href=""
+        imageAlt="a test image"
+        imageSrc="path-to-image.jpg"
+        tag=""
+        text=""
+        title=""
+      />,
+    );
```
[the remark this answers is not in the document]
rssilva: Because the way the types are being [used so far](https://github.com/Localitos/pluto/pull/1209/files#diff-f7fa8c6df160defa8d1aad22a1a3e5af7da525824cca21b138dee5363ceedbd9R20) when there is a href, both linkHref and linkText should be undefined and vice-versa this is not so easily done

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { Icon } from "../Icon";
+import { Button } from "../Button";
+import { Anchor } from "../Anchor";
+
+type CommonProps = {
+  imageSrc: string;
+  imageAlt: string;
+  title: string;
+  text: string;
+  tag?: string;
```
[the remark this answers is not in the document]
rssilva: on it!

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,232 @@
...
+  imagePosition?: ImagePosition;
+  iconUrl?: string;
+  background?: Background;
+};
+
+type InteractiveCard = {
+  href: string;
+  linkHref?: undefined;
+  linkText?: undefined;
+  as?: undefined;
```
apvale, line 31: Shouldn't the optional parameters be already undefined? Besides that, shouldn't these values be strings?

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,232 @@
...
+};
+
+type RegularCard = {
+  href?: undefined;
+  linkHref?: string;
+  linkText?: string;
+  as?: React.ComponentProps<typeof Box.div>["as"];
```
apvale, line 38: Thanks for that! ❤️

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,232 @@
...
+type InteractiveCard = {
+  href: string;
+  linkHref?: undefined;
+  linkText?: undefined;
+  as?: undefined;
+};
+
+type RegularCard = {
+  href?: undefined;
+  linkHref?: string;
```
apvale, line 36: Isn't it the opposite? The Interactive have "clickable" links? 🤔

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,232 @@
...
+  /** Sets a callback to content button */
+  onClickButton,
+  /** Sets an icon to be rendered over the image */
+  iconUrl,
+  /** Sets the image position on the card */
+  imagePosition = "right",
+  /** Sets the background color according with the type */
+  background = "default",
+  /** Used by StyledComponents to render the component as a specific tag. If href is passed it'll be rendered as "a" */
+  as = "div",
```
apvale, line 79: ❤️

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,232 @@
...
+            display="flex"
+            gap="space30"
+            h="56px"
+            justifyContent="center"
+            left="16px"
+            position={"absolute"}
+            top={16}
+            w="56px"
+          >
+            <Box.div h="24px" position="relative" w="24px">
```
apvale, line 216: Why not use the "space70" token?

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -0,0 +1,232 @@
...
+            borderRadius="borderRadiusCircle"
+            data-testid="service-content-card-icon"
+            display="flex"
+            gap="space30"
+            h="56px"
+            justifyContent="center"
+            left="16px"
+            position={"absolute"}
+            top={16}
+            w="56px"
```
apvale, line 214: Why not using the tokens here?

packages/components/src/components/ContentCard/ContentCard.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+const defaultProps = {
+  title: "The card title",
+  text: "Lorem ipsum dolor sit, consectetur. Aliquam ac a. Ligula at et, sodales vel purus.",
+  date: "Mon, Sept 5 2022",
+  tag: "🏠 Permanent Housing",
+  buttonText: "Go to Page",
+  imageSrc: "/images/beach-seal-rocks.jpg",
+};
+
+export const Default = Template.bind({});
```
apvale, line 25: What do you think about adding one more example rendering more than 1 card? Does it make sense? 🤔

apvale (approved):

packages/components/src/components/ContentCard/ContentCard.tsx
rssilva: `h` and `w` does not translate tokens like other props. When we use a token this is the result:

Screen Shot 2024-01-05 at 18 03 55

rssilva: I'm using numbers though

packages/components/src/components/ContentCard/ContentCard.tsx
rssilva: same as https://github.com/Localitos/pluto/pull/1209#discussion_r1443237021

packages/components/src/components/ContentCard/ContentCard.tsx
rssilva: I got that the other way. I'll double check before merging it

packages/components/src/components/ContentCard/ContentCard.tsx
rssilva: This is purely to make sure that we are not forced to pass a prop that is undefined

PR #1272, opened 2024-02-05
feat: accepts children prop to add a Button with composition

apvale:
## Description of the change

[ContentCard]:

- Accepts `children` so that it is possible to add a button flexibly
- Fixes a bug with the shadow state that should be shown in hover only when
  the card is clickable
  
 

https://github.com/Localitos/pluto/assets/5320963/fd1132a7-4220-4034-a08c-ecdd4371838e

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/ContentCard/ContentCard.tsx
+++ b/packages/components/src/components/ContentCard/ContentCard.tsx
@@ -3,7 +3,6 @@ import { Box, BoxProps } from "../../primitives/Box";
 import { Heading } from "../Heading";
 import { Text } from "../../primitives/Text";
 import { Icon } from "../Icon";
-import { Button } from "../Button";
 import { Anchor } from "../Anchor";
 
 type ImagePosition = "right" | "top";
@@ -23,18 +22,14 @@ type CommonProps = {
   tag?: string;
   /** Sets the date to be rendered together with calendar icon */
   date?: string;
-  /** Sets the content button text */
-  buttonText?: string;
-  /** Sets the content button as a link and adds target="_blank" and rel="noopener noreferrer" */
-  buttonAsLink?: boolean;
-  /** Sets a callback to content button */
-  onClickButton?: () => void;
   /** Sets the image position on the card */
   imagePosition?: ImagePosition;
   /** Sets an icon to be rendered over the image */
   iconUrl?: string;
   /** Sets the background color according with the type */
   background?: Background;
+  /** Accepts a children element to be rendered as the Card's content Button */
+  children?: React.ReactNode;
 };
 
 type InteractiveCard = {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -23,38 +23,26 @@ type CommonProps = {
...
+  /** Sets the href to be added to the interactive element */
+  href?: string;
+  /** Where to display the linked URL. Same as: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target */
+  target?: HTMLAnchorElement["target"];
+  /** Callback to be used when the interactive element is clicked */
+  onClick?: React.MouseEventHandler;
+  /** Sets the text to the interactive element when it's anchor or button */
+  ctaText?: string;
   /** Used by StyledComponents to render the component as a specific tag. If href is passed it'll be rendered as "a" */
   as?: React.ComponentProps<typeof Box.div>["as"];
```
nyan07, line 43: I'm not sure if we should keep this property because the developer doesn't have full control of it. And thinking about it, what would be the scenario to not use a `div` or a `a`?

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -103,28 +91,43 @@ export const ContentCard = ({
     },
   };
 
+  const isCardInteractive = interactiveElementType === "card";
+  const isButton = interactiveElementType === "button";
+  const inAnchor = interactiveElementType === "anchor";
```
nyan07, line 96: Maybe having `interractive` in all variables here?
Another thing I think would be good is using an enum. So you could use its value instead of strings. 

```typescript
isInteractiveElementType(InteractiveElementType.BUTTON);
```

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -103,28 +91,43 @@ export const ContentCard = ({
...
-      }}
-      cursor={href ? "pointer" : "default"}
+      boxShadow={
+        isCardInteractive
+          ? {
+              _: "none",
+              hover: "shadowStrong",
+            }
+          : "none"
+      }
```
nyan07, line 119:
```suggestion
      boxShadow={
        _: "none",
        hover: isCardInteractive && "shadowStrong",
      }
```

Not sure if it works.

nyan07 (approved):

packages/components/src/components/ContentCard/ContentCard.tsx
apvale: It will be used when the dev wants to create a listitem because we can list more than one card.

PR #1350, opened 2024-02-29
[NOISSUE] Add GitHub Workflow to remind about too many open PRs

pecavalheiro:
### Summary
The same script that being used in the backend repo.

### What kind of change is this
- _Non-breaking_ change: Internal improvement or bug fixing


padulafacundo (commented):
Can we change the commit so it follows [conventional commits](https://github.com/Localitos/pluto/blob/main/CONTRIBUTING.md#conventional-commits)? Also, do we need a changeset?

pecavalheiro:
@padulafacundo I've adjusted the commit, please let me know if more details are necessary 😊 

About the changeset, I'm not familiar with how it's done here. From what I understand, I should either update or delete the old file. Is that correct or are you able to point me on how to proceed?

rssilva (approved):

rssilva:
I believe we don't need a changeset since this changes are not packed

PR #1351, opened 2024-02-29
feat: add border and background color white to Table component

apvale:
## Description of the change

This PR will add a gray outline and a wrapper with a background color white to the Table component.

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Table/Table.tsx
+++ b/packages/components/src/components/Table/Table.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { styled } from "@localyze-pluto/theme";
+import { SystemProp, Theme, styled } from "@localyze-pluto/theme";
 import PropTypes from "prop-types";
 import { Box } from "../../primitives/Box";
 import { TableContext } from "./TableContext";
@@ -14,6 +14,8 @@ export interface TableProps
   striped?: boolean;
   /** Determines how to layout the cells, rows, and columns. */
   tableLayout?: "auto" | "fixed";
+  /** The background color of the table wrapper. */
+  backgroundColor?: SystemProp<keyof Theme["colors"], Theme>;
 }
 
 const StyledTable = styled(Box.table)<TableProps>`
@@ -28,6 +30,7 @@ const Table = React.forwardRef<HTMLTableElement, TableProps>(
       children,
       striped = false,
       tableLayout = "auto",
+      backgroundColor = "bgPrimary",
       ...props
     },
     ref,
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Table/Table.tsx
```diff
@@ -36,15 +36,25 @@ const Table = React.forwardRef<HTMLTableElement, TableProps>(
...
+          h="100%"
+          overflow="scroll"
         >
-          {children}
-        </StyledTable>
+          <StyledTable
+            borderCollapse="separate"
+            ref={ref}
+            tableLayout={tableLayout}
+            w="100%"
```
rssilva, line 52: what if we add a backgroundColor property to the component, adding `bgPrimary` (new token who dis 🎉) as default? So it would still be possible to change it and the change would still be made?

.changeset/sixty-feet-try.md
```diff
@@ -0,0 +1,7 @@
+---
+"@localyze-pluto/components": patch
```
padulafacundo, line 2: according to [semantic versioning](https://semver.org/), a patch is a backward compatible bugfix.
I would say this is either a breaking change (because it changes the behavior of an existing component) or a minor change. And considering that semantic version explicitly says that a breaking change is a change with incompatible API changes, then I guess it should be a minor change. But I', not sure if we ever officially decided to follow that

padulafacundo (approved):

packages/components/src/components/Table/Table.tsx
```diff
@@ -36,15 +36,25 @@ const Table = React.forwardRef<HTMLTableElement, TableProps>(
 
     return (
       <TableContext.Provider value={tableContext}>
-        <StyledTable
-          borderCollapse="separate"
-          ref={ref}
-          tableLayout={tableLayout}
-          w="100%"
-          {...props}
+        <Box.div
```
rssilva, line 39: The reason to put this wrapper is the border and background, right? What if we are capable to add these styles not adding a div to it? My concern here is to wrap the table around an element to solve something that could be made by pure CSS

padulafacundo: Could we put it in the styled table?

.changeset/sixty-feet-try.md
apvale: Changed to minor.

rssilva (approved):

packages/components/src/components/Table/Table.tsx
rssilva: Ana showed me the troubles with sticky headers not having the wrapper and life sometimes can be hard

apvale: It doesn't work as expected if we add to the StyledTable just because of the Sticky variant. :( 

In the video, you can see that when I scroll down, the table goes up and therefore the border top and bottom disappear because they are part of the Table. That's why I added the wrapper. 
https://github.com/Localitos/pluto/assets/5320963/ec8fc0c3-93bc-44d5-95aa-766904461362

packages/components/src/components/Table/Table.tsx
apvale: Changed! Thank you :)

PR #1378, opened 2024-03-06
Add `flexed` option to `Tabs` component

sliminas:
## Description of the change

For the new reporting graphs we will add tabs to the reporting page which breaks the AG Grid results as they use 100% height for the result set, which leads to no results at all being visible since the Tabs wrapper div has display block and therefore does not fill the whole view port height.

Screenshot_20240306_092242
Screenshot_20240306_092306

## Type of change

### Before (without `flexed`)
Screenshot_20240306_095618

### After (with `flexed`)
Screenshot_20240306_095635

### DIFF
--- a/packages/components/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/components/src/components/Tabs/Tabs.stories.tsx
@@ -316,7 +316,7 @@ export const Flexed = (): React.JSX.Element => {
           </Tab>
         </TabList>
         <Box.div display="flex" flexDirection="column" flexGrow="1">
-          {selectedTab === "tab1" ? <Tab1/> : <Tab2/>}
+          {selectedTab === "tab1" ? <Tab1 /> : <Tab2 />}
         </Box.div>
       </Tabs>
     </Box.div>
--- a/packages/components/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/components/src/components/Tabs/Tabs.stories.tsx
@@ -277,38 +277,46 @@ WithDivider.parameters = {
   },
 };
 
-const Foo = () => (
+const Tab1 = () => (
   <Box.div
-    borderColor="colorBackgroundComplete"
+    borderColor="colorBorderWeaker"
     borderStyle="borderStyleSolid"
     borderWidth="borderWidth10"
     h="100%"
+    padding="space60"
     w="100%"
   >
-    <h3>Foo</h3>
-    Lorem ipsum dolor sit, consectetur. Aliquam ac a. Ligula at et, sodales vel
-    purus.
+    Tab1
   </Box.div>
 );
 
-const Bar = () => <div>Bar</div>;
+const Tab2 = () => (
+  <Box.div
+    borderColor="colorBorderWeaker"
+    borderStyle="borderStyleSolid"
+    borderWidth="borderWidth10"
+    padding="space60"
+  >
+    Tab2
+  </Box.div>
+);
 
 export const Flexed = (): React.JSX.Element => {
-  const [selectedTab, setSelectedTab] = React.useState("foo");
+  const [selectedTab, setSelectedTab] = React.useState("tab1");
 
   return (
     <Box.div display="flex" h="500px">
       <Tabs flexed>
         <TabList aria-label="Page tabs" withDivider>
-          <Tab id="foo" onClick={() => setSelectedTab("foo")}>
-            Foo
+          <Tab id="tab1" onClick={() => setSelectedTab("tab1")}>
+            Tab1
           </Tab>
-          <Tab id="bar" onClick={() => setSelectedTab("bar")}>
-            Foo
+          <Tab id="tab2" onClick={() => setSelectedTab("tab2")}>
+            Tab2
           </Tab>
         </TabList>
         <Box.div display="flex" flexDirection="column" flexGrow="1">
-          {selectedTab === "foo" ? <Foo /> : <Bar />}
+          {selectedTab === "tab1" ? <Tab1/> : <Tab2/>}
         </Box.div>
       </Tabs>
     </Box.div>
--- a/packages/components/src/components/Tabs/Tabs.stories.tsx
+++ b/packages/components/src/components/Tabs/Tabs.stories.tsx
@@ -277,33 +277,38 @@ WithDivider.parameters = {
   },
 };
 
+const Foo = () => (
+  <Box.div
+    borderColor="colorBackgroundComplete"
+    borderStyle="borderStyleSolid"
+    borderWidth="borderWidth10"
+    h="100%"
+    w="100%"
+  >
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,45 @@ WithDivider.parameters = {
     },
   },
 };
+
+export const Flexed = (): React.JSX.Element => {
```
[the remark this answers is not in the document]
rssilva: It's missing the `<TabPanels>` and `<TabPanel>`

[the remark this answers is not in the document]
sliminas: Here is an example of the intended use case @rssilva @padulafacundo 
https://github.com/Localitos/Backend/blob/d926381ac819200fe8f20380a084c4c3b44af548/client/src/deprecated/pages/viewsNew/ChewpacaLayout/TalentProfile/components/TalentCommunication/TalentCommunication.tsx#L61-L79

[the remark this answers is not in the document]
sliminas: I made the tabs in the story work in daedfb0

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,50 @@ WithDivider.parameters = {
...
+        <TabList aria-label="Page tabs" withDivider>
+          <Tab id="foo" onClick={() => setSelectedTab("foo")}>
+            Foo
+          </Tab>
+          <Tab id="bar" onClick={() => setSelectedTab("bar")}>
+            Foo
+          </Tab>
+        </TabList>
+        <Box.div display="flex" flexDirection="column" flexGrow="1">
+          {selectedTab === "foo" ? <Foo /> : <Bar />}
```
apvale, line 311: You don't need to do this if you use the TabPanels component and have an initialTabId in the `<Tabs />` component. To keep it consistent and use the entire power the Tabs component provides us, we need to use the `TabPanels` component. In this case:

```
<TabPanels>
  <TabPanel>
    <Foo />
  </TabPanel>
  <TabPanel>
    <Bar />
  </TabPanel>
/TabPanels>
```
In this way, you don't need to set an onClick to the Tab component.

apvale (changes requested):
I propose to use the TabPanels in the stories since we avoid adding extra behavior (when the component already covers it) and avoid changing how the Tabs were supposed to be used.

packages/components/src/components/Tabs/Tabs.stories.tsx
sliminas: @apvale :point_up:

apvale: Good point. I hadn't thought about that. Two things:

1. Pluto stories should be generic and cover the scenario you are adding, so, the flexed tab. So, for me, you can use the TabPanels and TabPanel here to exemplify the new prop. 
2. The issue you mentioned above needs to be fixed in a different way by adding a new prop, for example, to render a TabPanel content only when tab is active. 

For the scenario you mentioned, you can continue to use TabPanels and render the TabPanel content in a certain condition. Like:

```
<TabPanels>
  <TabPanel>
    {condition && <Component />}
  </TabPanel>
</TabPanels>
```

apvale: ChakraUI has a good example of it that we can try to do for Pluto: https://chakra-ui.com/docs/components/tabs#lazily-mounting-tab-panels

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,50 @@ WithDivider.parameters = {
...
+
+  return (
+    <Box.div display="flex" h="500px">
+      <Tabs flexed>
+        <TabList aria-label="Page tabs" withDivider>
+          <Tab id="foo" onClick={() => setSelectedTab("foo")}>
+            Foo
+          </Tab>
+          <Tab id="bar" onClick={() => setSelectedTab("bar")}>
+            Foo
```
rssilva, line 307:
```suggestion
            Bar
```

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,50 @@ WithDivider.parameters = {
...
+    h="100%"
+    w="100%"
+  >
+    <h3>Foo</h3>
+    Lorem ipsum dolor sit, consectetur. Aliquam ac a. Ligula at et, sodales vel
+    purus.
+  </Box.div>
+);
+
+const Bar = () => <div>Bar</div>;
```
rssilva, line 294:
```suggestion
const Bar = () => (
  <Box.div padding="space60">
    <Heading as="h3" marginBottom="space0" size="heading50">
      Bar
    </Heading>
  </Box.div>
);
```

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,50 @@ WithDivider.parameters = {
...
+
+const Foo = () => (
+  <Box.div
+    borderColor="colorBackgroundComplete"
+    borderStyle="borderStyleSolid"
+    borderWidth="borderWidth10"
+    h="100%"
+    w="100%"
+  >
+    <h3>Foo</h3>
```
rssilva, line 288:
```suggestion
    <Heading as="h3" marginBottom="space0" size="heading50">
      Foo
    </Heading>
```

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,50 @@ WithDivider.parameters = {
     },
   },
 };
+
+const Foo = () => (
+  <Box.div
+    borderColor="colorBackgroundComplete"
```
rssilva, line 282:
```suggestion
    borderColor="colorBorderWeaker"
```

packages/components/src/components/Tabs/Tabs.stories.tsx
```diff
@@ -276,3 +276,50 @@ WithDivider.parameters = {
...
+    borderStyle="borderStyleSolid"
+    borderWidth="borderWidth10"
+    h="100%"
```
rssilva, line 285:
```suggestion
    h="100%"
    padding="space60"
```

rssilva (approved):
I have suggestions regarding the stories UI but not blockers

packages/components/src/components/Tabs/Tabs.stories.tsx
sliminas: Why adding so much code for a story for an element that is not the focus of the story?
Isn't this distracting from the actual use case?

PR #1385, opened 2024-03-12
feat: add isHoverable prop to table row

apvale:
## Description of the change

When we updated the Table component to unset the hover state when a row wasn't clickable, we broke the striped version, when isClickable and striped could not work at the same time. 

This PR adds a new property called `isHoverable` to the table row and makes it possible to be hoverable and striped at the same time. The `isClickable` prop now only changes the `cursor` to `pointer` when true. 

Besides that, I put it back the old background tokens because we need to first introduce the specific tokens like "bgHover" and then update them in places like tables.

All the examples can be seen [here](https://632b1baf1257171b6a84f370-nsoxbxopty.chromatic.com/?path=/docs/components-table--docs)

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Table/Table.stories.tsx
+++ b/packages/components/src/components/Table/Table.stories.tsx
@@ -143,7 +143,7 @@ export const StickyHeaders: Story = {
   ),
 };
 
-export const WithClickableRows: Story = {
+export const WithClickableRowsAndHoverable: Story = {
   render: () => (
     <Box.div h="20rem" overflowY="auto">
       <Table>
@@ -156,7 +156,7 @@ export const WithClickableRows: Story = {
         </THead>
         <TBody>
           {map([...Array.from({ length: 100 }).keys()], (index) => (
-            <Tr isClickable key={index}>
+            <Tr hasHover isClickable key={index}>
               <Td>Content</Td>
               <Td>Content</Td>
               <Td>Content</Td>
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Table/Table.stories.tsx
```diff
@@ -156,7 +156,7 @@ export const WithClickableRows: Story = {
         </THead>
         <TBody>
           {map([...Array.from({ length: 100 }).keys()], (index) => (
-            <Tr isClickable key={index}>
+            <Tr isClickable isHoverable key={index}>
```
rssilva, line 159: I believe that `isHoverable` does not translate well the fact that there is an interaction on hover because technically every row is hoverable (some has interaction, some does not)

apvale: So, what could be better? Adding a prop to the table and using it from context to change the hover of the row?

rssilva: just the name of the prop if anyone else has an idea `hasHover`?

apvale: Thank you! Updated it here https://github.com/Localitos/pluto/pull/1385/files#diff-6b7a9252166d5be1cc435ed1af0e1469b9ef3739cbee9d45b69b15accfcdd1a5R16

rssilva (approved):
🔲 🟫 🔲 🔲 🔲 (this is my lame representation of hovering)

Please remember to change the commit description when squashing!

PR #1484, opened 2024-04-03
feat: add backdrop to drawer component

apvale:
## Description of the change

This PR adds a styled backdrop to the Drawer component. This component could use the `data-backdrop` attribute already present in `base.tsx` styles, but I decided not to change that right now to avoid impacting other components like Modal. 

Besides that, it fixes an issue with the initialFocus that was focusing the `X` button when the Drawer was opened.

https://github.com/Localitos/pluto/assets/5320963/350032b3-ff20-401e-82ae-582636f3efca

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/.changeset/red-chefs-sin.md
+++ b/.changeset/red-chefs-sin.md
@@ -1,0 +1,9 @@
+---
+"@localyze-pluto/components": minor
+---
+
+[Drawer]:
+
+- Adds backdrop to Drawer component
+- Control with autoFocusOnShow prop whether the Drawer
+  should focus the first focusable element when shown
--- a/packages/components/src/components/Drawer/Drawer.tsx
+++ b/packages/components/src/components/Drawer/Drawer.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { Dialog } from "ariakit/dialog";
 import type { DialogProps } from "ariakit";
+import { styled } from "@localyze-pluto/theme";
 import { Box } from "../../primitives/Box";
 
 export interface DrawerProps extends DialogProps {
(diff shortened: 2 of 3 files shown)


packages/components/src/components/Drawer/Drawer.tsx
```diff
@@ -1,25 +1,33 @@
...
 /** A Drawer is a page overlay that displays information and blocks interaction with the page until an action is taken or the Drawer is dismissed. */
 const Drawer = React.forwardRef<HTMLDivElement, DrawerProps>(
-  ({ children, ...props }, ref) => {
+  ({ children, initialFocusRef, ...props }, ref) => {
     return (
       <Box.div
         as={Dialog}
-        backdrop={false}
+        autoFocusOnShow={initialFocusRef ?? false}
+        backdrop={StyledBackdrop}
```
padulafacundo, line 24: > - Adds backdrop prop to Drawer component

from the description of the change I thought this would be an optional prop. Are we always going to have the backdrop? If so maybe we can update the message since it's actually not a prop of the component, but a prop that we pass to the div inside this component

> - Adds backdrop ~prop~ to Drawer component

.changeset/red-chefs-sin.md
```diff
@@ -0,0 +1,8 @@
+---
+"@localyze-pluto/components": patch
+---
+
+[Drawer]:
+
+- Adds backdrop prop to Drawer component
+- Fixes initialFocus prop to focus element only when necessary
```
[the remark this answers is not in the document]
apvale: I think it's not necessary since this component was never used. But I can create just one minor and consider both things as new changes.

packages/components/src/components/Drawer/Drawer.tsx
apvale: So you think the issue is with the changelog message?

padulafacundo: yeah maybe, when I read the message I thought it was going to be optional (that we could control it with a prop, and set the backdrop on or off)

padulafacundo (approved):

PR #1515, opened 2024-04-11
fix: avoid click on button to close

rssilva:
## Description of the change
- Change the place where the Drawer test has a click in order to fix flakiness

## Type of change

### DIFF
--- a/packages/components/src/components/Drawer/Drawer.test.tsx
+++ b/packages/components/src/components/Drawer/Drawer.test.tsx
@@ -56,7 +56,7 @@ describe("<Drawer />", () => {
     expect(await screen.findByRole("dialog")).toBeInTheDocument();
     expect(screen.getByRole("heading", { level: 2 })).toBeInTheDocument();
 
-    await user.click(renderedOpenButton);
+    await user.click(await screen.findByRole("presentation"));
     await waitFor(() => {
       expect(screen.getByRole("dialog")).not.toBeVisible();
     });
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Drawer/Drawer.test.tsx
```diff
@@ -48,14 +48,15 @@ describe("<Drawer />", () => {
...
+
     const renderedOpenButton = screen.getByText(OPEN_DRAWER_TEXT);
     expect(renderedOpenButton).toBeInTheDocument();
 
     await user.click(renderedOpenButton);
     expect(await screen.findByRole("dialog")).toBeInTheDocument();
     expect(screen.getByRole("heading", { level: 2 })).toBeInTheDocument();
 
-    await user.click(renderedOpenButton);
+    await user.click(await screen.findByRole("presentation"));
```
apvale, line 59: Why the issue it's related to the place is clicked? The flakiness is also visible in the test to check the `Escape` key.

rssilva: we can remove the animations while running tests to check if that is also affecting it but firing a click on a button that should be hidden on the interface doesn't seem a good test

apvale: You have a good point about this test. I didn't notice that this click is on the button that is not visible. So, for this test, it makes sense to click on the presentation.

apvale (approved):

PR #1490, opened 2024-04-05
[NO ISSUE] feat: allow Drawer to receive padding as prop

helfer1991:
## Description of the change

Goal of the PR is to allow the Drawer component to receive padding as a Prop, therefore increasing its flexibility and allowing to meet the design goals stated here, where padding is 24px (space70).

## Testing the change

- [ ] Write your testing instructions here

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Non-Breaking Change (change to existing functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Development

- [ ] The code changed/added as part of this pull request has been covered with tests
- [ ] All tests related to the changed code pass in development

### Code review

- [ ] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached.

PR #1519, opened 2024-04-11
[NO ISSUE] feat: change fontSize of drawer heading

helfer1991:
## Description of the change

Goal of the PR is to make the Drawer component more flexible, allowing, for example, the title/heading to have a marginBottom and to fix its font-size to match the designs for the Partners user story of the TREX team
## Testing the change

## Type of change

### Development

### Code review


packages/components/src/components/Drawer/DrawerHeader.tsx
```diff
@@ -1,23 +1,30 @@
...
+    { children, marginBottom = "space0", padding = "space60", ...props },
+    ref,
+  ) => {
     return (
       <Box.div
+        alignItems="center"
         display="flex"
         gap="space30"
         justifyContent="space-between"
+        marginBottom={marginBottom}
```
apvale, line 27: Why do we need a marginBottom? I ran storybook and noticed that it's showing ok for what we need, no? I mean, padding should be enough. Also, we don't set marginBottom on Figma.

image

packages/components/src/components/Heading/Heading.tsx
```diff
@@ -6,7 +6,7 @@ import forEach from "lodash/forEach";
 import { Text } from "../../primitives/Text";
 
 type HeadingLevelOptions = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
-type HeadingMarginOptions = "space0" | "space70";
+type HeadingMarginOptions = SystemProp<keyof Theme["space"], Theme>;
```
apvale, line 9: Why did you change this?

helfer1991: well, because instead of `"space0" | "space60" | "space160" | ...` I think it's more appropriate to make this more flexible and able to receive any `spaceSuffix` available

packages/components/src/components/Drawer/DrawerHeader.tsx
helfer1991: I get your point completely, this is just to provide more flexibility to this component for future use.
If you think this specific change should be deleted I'm fine with it, just let me know.

packages/components/src/components/Heading/Heading.tsx
apvale: I understand the reason to change this, but, IMHO, it shouldn't be part of this PR. This involves changing the Heading component that's not even being mentioned in the changeset, for example.

packages/components/src/components/Drawer/DrawerHeader.tsx
apvale: For me, with the changes you did on this [PR](https://github.com/Localitos/pluto/pull/1490), we already have a padding of 16px that it's the same what we have on Figma. So, it seems to me that we don't need to introduce anything related to spaces on this PR.

packages/components/src/components/Heading/Heading.tsx
helfer1991: yeah, you're right. reverted those changes

.changeset/wise-feet-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"@localyze-pluto/components": minor
+---
+
+enable margin bottom to the title and change its font size
```
apvale, line 5: I think we just want to mention that we update the fontSize, right?

apvale (approved):

.changeset/wise-feet-clean.md
helfer1991: yes, forgot about this, good catch

helfer1991: done

PR #1590, opened 2024-05-13
fix(components): add button with 100% width correctly to menu

rssilva:
## Description of the change
- Add 100% width to the button on menu avoiding to be overwritten by previous button style

background added for pure debugging

before:
Screen Shot 2024-05-13 at 08 00 20

after:
Screen Shot 2024-05-13 at 08 00 06

## Type of change

### Development

### DIFF
--- a/.changeset/neat-wasps-yell.md
+++ b/.changeset/neat-wasps-yell.md
@@ -1,0 +1,5 @@
+---
+"@localyze-pluto/components": minor
+---
+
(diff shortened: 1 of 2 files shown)


sliminas:
@rssilva Can't we make it fill the whole area? After that change there will still be a big border that is not clickable.

rssilva:
@sliminas it's clickable! The button has padding but it'll trigger the event

helfer1991 (approved):

PR #1615, opened 2024-06-05
chore: add component tokens

rssilva:
## Description of the change

Add component tokens

## Type of change


helfer1991 (approved):

.changeset/great-flies-yawn.md
```diff
@@ -0,0 +1,6 @@
+---
```
apvale, line 1: This one can deleted, right?

apvale (commented):
Code looks good to me! Thanks, Rafa 💟 

I ran the storybook and saw that the RGB values are broken in the colors list. Do you know why? 

image

rssilva:
@apvale good catch! I haven't noticed 🤦 

This happens because the docs used to get the colors straight from the colors JSON directly and that was an hexadecimal value. Now it's a template variable.

https://github.com/Localitos/pluto/blob/main/packages/design-tokens/docs/components/constants.tsx#L48

apvale (approved):

PR #1631, opened 2024-06-13
chore(design-tokens) chore: add d tokens for space

rssilva:
## Description of the change

- Add `d` tokens for space in order to replace `space` prefix in the future


packages/design-tokens/config.js
```diff
@@ -234,20 +250,22 @@ module.exports = {
         },
         {
           destination: "space.js",
-          format: "javascript/es6",
-          filter: {
-            attributes: {
-              category: "space",
-            },
+          format: "jsPrefixFormatter",
```
apvale, line 253: We are still using the `space` prefix, right? But I couldn't see them in the `spacing` list in Storybook. 🤔 Are we already removing them from there?

apvale (approved):

packages/design-tokens/config.js
rssilva: Oh crap! I forgot to push the fix to the docs 🤦 

https://github.com/Localitos/pluto/pull/1631/files#diff-f3a29ad55c2b1ac357952013816bd161f3914b9422efb7a17385975bf5704ed9R111

PR #1675, opened 2024-07-17
chore(components):  remove tokens limitations

rssilva:
## Description of the change

- Remove token limitations from components below to make them easily changeable 

```
DrawerHeader
Heading
Label
OrderedList
UnorderedList
ModalBody
Paragraph
Td
```

## Type of change

### DIFF
--- a/packages/components/src/components/Drawer/DrawerHeader.tsx
+++ b/packages/components/src/components/Drawer/DrawerHeader.tsx
@@ -1,25 +1,23 @@
 import React from "react";
 import { DialogDismiss } from "@ariakit/react/dialog";
-import { Box } from "../../primitives/Box";
+import { Box, BoxProps } from "../../primitives/Box";
 import { Button } from "../Button";
 
-export interface DrawerHeaderProps
-  extends Omit<React.HTMLAttributes<HTMLDivElement>, "color"> {
+export interface DrawerHeaderProps extends BoxProps {
   /** The contents of the drawer header */
   children: NonNullable<React.ReactNode>;
-  padding?: "p0" | "p5" | "space0" | "space60";
 }
 
 /** The header content area of the drawer. */
 const DrawerHeader = React.forwardRef<HTMLDivElement, DrawerHeaderProps>(
-  ({ children, padding = "space60", ...props }, ref) => {
+  ({ children, ...props }, ref) => {
     return (
       <Box.div
         alignItems="center"
         display="flex"
         gap="space30"
         justifyContent="space-between"
-        padding={padding}
+        padding="space60"
         ref={ref}
         {...props}
       >
--- a/packages/components/src/components/Heading/Heading.tsx
+++ b/packages/components/src/components/Heading/Heading.tsx
@@ -3,10 +3,9 @@ import type { SystemProp, Theme } from "@xstyled/styled-components";
 import toPairs from "lodash/toPairs";
 import isObject from "lodash/isObject";
 import forEach from "lodash/forEach";
-import { Text } from "../../primitives/Text";
+import { Box, BoxProps } from "../../primitives/Box";
 
 type HeadingLevelOptions = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
-type HeadingMarginOptions = "m0" | "m6" | "space0" | "space70";
 type HeadingSizeOptions =
   | "heading10"
   | "heading20"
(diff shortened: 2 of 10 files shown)


alpacatron:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 83a4df791e6905420c7d20d6fba8a32f0e3d40f4
</p>

packages/components/src/components/Heading/Heading.tsx
```diff
@@ -29,25 +28,15 @@ type HeadingSizeBreakpoints = {
...
+  extends Omit<React.HTMLAttributes<HTMLHeadingElement>, "color">,
+    BoxProps {
   /** Sets the HTML element on render. */
   as?: HeadingLevelOptions;
   /** The contents of the heading. Can be text or valid text related HTML, i.e. anchor and strong elements. */
   children: NonNullable<React.ReactNode>;
-  /** Changes the bottom margin of the heading.  */
-  marginBottom?: HeadingMarginOptions;
   /** Changes the size of the heading. */
   size?: HeadingSizeOptionsProp;
```
apvale, line 39: If we pass the prop `color` here as BoxProps["color"], we don't need to cast it in here https://github.com/Localitos/pluto/pull/1675/files?diff=split&w=1#diff-79d1fad6284384507b904515127be29ac7ca2e1a6d2ffb387c1fc8af0c9fa7e1R149.

packages/components/src/components/Label/Label.tsx
```diff
@@ -31,7 +25,7 @@ const Label = React.forwardRef<HTMLLabelElement, LabelProps>(
         fontSize="fontSize10"
         fontWeight="fontWeightMedium"
         lineHeight="lineHeight10"
-        marginBottom={marginBottom}
+        marginBottom="space20"
         ref={ref}
         {...props}
```
apvale, line 30: So, if we pass another marginBottom to `Label` component while using it, it will override this "space20", right?

packages/components/src/components/Label/Label.tsx
```diff
@@ -31,7 +25,7 @@ const Label = React.forwardRef<HTMLLabelElement, LabelProps>(
         fontSize="fontSize10"
         fontWeight="fontWeightMedium"
         lineHeight="lineHeight10"
-        marginBottom={marginBottom}
+        marginBottom="space20"
```
apvale, line 28: Shouldn't we the new tokens here?

packages/components/src/components/Drawer/DrawerHeader.tsx
```diff
@@ -1,25 +1,23 @@
...
-  ({ children, padding = "space60", ...props }, ref) => {
+  ({ children, ...props }, ref) => {
     return (
       <Box.div
         alignItems="center"
         display="flex"
         gap="space30"
         justifyContent="space-between"
-        padding={padding}
+        padding="space60"
```
apvale, line 20: Can we use the new token here?

packages/components/src/components/Paragraph/Paragraph.tsx
```diff
@@ -45,20 +41,20 @@ const getParagraphStyles = (
 
 /** A Paragraph is a block of text. */
 const Paragraph = React.forwardRef<HTMLParagraphElement, ParagraphProps>(
-  ({ children, marginBottom = "space70", size = "medium", ...props }, ref) => {
+  ({ children, size = "medium", ...props }, ref) => {
     return (
-      <Text.p
+      <Box.p
```
apvale, line 46: Do you know the difference between using Text or Box?
Why can't we continue using Text in here?

apvale (approved):

packages/components/src/components/Heading/Heading.tsx
rssilva: I got an error while doing it:

```
The expected type comes from property 'color' which is declared here on type 'IntrinsicAttributes & { col?: SystemProp<string | number | true, Theme> | undefined; p?: PaddingProp<Theme> | undefined; ... 440 more ...; listStylePosition?: SystemProp<...> | undefined; } & { ...; } & { ...; }'
```

packages/components/src/components/Label/Label.tsx
rssilva: yes!

packages/components/src/components/Label/Label.tsx
rssilva: we could but I preferred to avoid a diff that is harder to read

packages/components/src/components/Paragraph/Paragraph.tsx
rssilva: I got an error that comes from the type since it was accepting only text props (fontWeight, color) instead of accepting more props like margin, padding, etc

PR #1716, opened 2024-09-04
fix(components): Update menu outline color to gray

sliminas:
## Description of the change

The outline color of the menu is inconsistent with the rest of the border colors we use and e.g. also inconsistent with the current user menu we have in the classic app.

### Current menu outline
Screenshot_20240904_152315

### Old user menu in classic app with HeadlessUI menu
image

When replacing the old headlessUI menu with the Pluto menu the outline makes the borders look different than in all the other places.

### new user menu in classic app with Pluto menu
image

### DIFF
--- a/packages/components/src/components/Menu/Menu.tsx
+++ b/packages/components/src/components/Menu/Menu.tsx
@@ -63,7 +63,7 @@ const Menu = React.forwardRef<HTMLButtonElement, MenuProps>(
             flexDirection="column"
             maxWidth="280px"
             minWidth="160px"
-            outlineColor="colorBorderweaker"
+            outlineColor="colorBorderWeaker"
             outlineStyle="borderStyleSolid"
             outlineWidth="borderWidth10"
           >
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Menu/Menu.tsx
```diff
@@ -63,7 +63,7 @@ const Menu = React.forwardRef<HTMLButtonElement, MenuProps>(
             flexDirection="column"
             maxWidth="280px"
             minWidth="160px"
-            outlineColor="colorBorderPrimary"
+            outlineColor="colorBorderweaker"
```
padulafacundo, line 66: shouldn't this be 
```suggestion
            outlineColor="colorBorderWeaker"
```
?

Did you copy it from storybook? I think we have an issue with token names that have more than two words:

image

Also, I would have expected typescript to catch this?

sliminas: Yes I copied it from there. :slightly_smiling_face:

PR #1736, opened 2024-09-18
fix(docs): change TOKEN_COLUMNS to avoid conflicting tokens with the same name

rssilva:
## Description of the change

- Add a name property to TokensTable
- Add a null check to TokensTable to avoid breaking by accessing an undefined value
- Change TOKENS_TABLE to have one more level (or prefix) to each object. So a titleDisplay that comes from font-size would not collide with titleDisplay that comes from font-weight
- Call camelcase on tokens names avoiding things as `colorBackgrounddecorative-strong` correcting it to `colorBackgroundDecorativeStrong`

## Type of change

### BEFORE

Screen Shot 2024-09-18 at 08 50 40
Screen Shot 2024-09-18 at 08 50 31

### AFTER
Screen Shot 2024-09-18 at 08 50 22
Screen Shot 2024-09-18 at 08 50 48

### DIFF
--- a/packages/design-tokens/docs/components/TokensTable.tsx
+++ b/packages/design-tokens/docs/components/TokensTable.tsx
@@ -18,15 +18,17 @@ import { TokenName } from "./TokenName";
 
 export const TokensTable = ({
   data,
+  name,
 }: {
   data: Record<string, Record<string, Token>>;
+  name: string;
 }): JSX.Element => {
   const tokenNames: Array<string> = filter(
     keys(data),
     (item) => item !== "default",
   );
   const columnGroups = map(tokenNames, (token) => {
-    return TOKEN_COLUMNS[token];
+    return TOKEN_COLUMNS[name][token];
   });
   const rowGroups: TokenRow[][] = map(columnGroups, (columns, index) => {
     const token = tokenNames[index];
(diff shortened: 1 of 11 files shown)


packages/design-tokens/docs/components/constants.tsx
```diff
@@ -105,7 +107,7 @@ const FONT_SIZE = reduce(
         name: "Name",
         transform: getTokenNameFromTuple(prefix),
       },
-      { name: "Pixels", transform: getTokenComment },
+      { name: "Weight", transform: getTokenValue },
```
apvale, line 110: Shouldn't we use `Weight` only for font-weight? 

image

packages/design-tokens/docs/components/TokensTable.tsx
```diff
@@ -18,15 +18,17 @@ import { TokenName } from "./TokenName";
 
 export const TokensTable = ({
   data,
+  name,
```
apvale, line 21: I noticed that we are using `category` in the type. What do you think of also using it here? Or maybe changing there to `name` so we have the same naming in both places?

apvale (approved):
Thanks! 🙏🏻

packages/design-tokens/docs/components/constants.tsx
rssilva: I'm removing this column since it doesn't make sense on font size!

packages/design-tokens/docs/components/TokensTable.tsx
rssilva: good call! I'm using category

PR #1743, opened 2024-09-26
feat: Update default color ModalHeading DrawerHeading

franziiiini:
## Description of the change

After updating the `Heading` default color ( in https://github.com/Localitos/pluto/pull/1731), we realized that there are still headings using the prior color. With this change we should have address all headings and have the same color.

## Type of change

DrawerHeading
Screenshot 2024-09-26 at 15 06 33

ModalHeading
Screenshot 2024-09-26 at 15 06 58

### DIFF
--- a/packages/components/src/components/Drawer/DrawerHeading.tsx
+++ b/packages/components/src/components/Drawer/DrawerHeading.tsx
@@ -13,7 +13,7 @@ const DrawerHeading = React.forwardRef<HTMLHeadingElement, DrawerHeadingProps>(
     return (
       <Heading
         as="h2"
-        color="colorTextHeadingStronger"
+        color="contentPrimary"
         marginBottom="m0"
         ref={ref}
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Drawer/DrawerHeading.tsx
```diff
@@ -13,7 +13,7 @@ const DrawerHeading = React.forwardRef<HTMLHeadingElement, DrawerHeadingProps>(
     return (
       <Heading
         as="h2"
-        color="colorTextHeadingStronger"
+        color="contentPrimary"
```
apvale, line 16: I think we don't even need to pass the color, right?

franziiiini: you are absolutely right! I'll remove it!

apvale (approved):

PR #1752, opened 2024-10-22
feat: update ContentCard component

apvale:
## Description of the change

This PR:
* has a breaking change because it changes the `iconUrl` prop to `topLeftElement` since it will accept ReactNode instead of only string. This allows us to use different image components, like Image from nextjs or other type of elements
* adds a new `topRightElement` prop so we allow adding an element in the top right corner of the card
* adds a new `lowOpacity` prop to set the opacity of the card content without including the `topRightElement`
* fixes some style inconsistencies (heading, card content) between the component in Figma and the one from Pluto
* adds `h` and `w` properties to allow setting custom width and height for the card

Disclaimer: we don't have any usage of the `iconUrl`, but I decided to make it a breaking change because of semantics.

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/.changeset/cold-news-glow.md
+++ b/.changeset/cold-news-glow.md
@@ -1,0 +1,14 @@
+---
+"@localyze-pluto/components": major
+---
+
+[ContentCard]
+
+> Breaking change: `iconUrl` prop is now `topLeftElement` and accepts a ReactNode
+
+Minor changes:
+
+- Set Icon wrapper border radius to 8px (shows now like a square)
+- Added new `topRightElement` prop to allow add element in the top right corner of the card
+- Added new `lowOpacity` prop to set the opacity of the card content without include the `topRightElement`
+- Fixed styles: heading color, card content
--- a/packages/components/src/components/ContentCard/ContentCard.tsx
+++ b/packages/components/src/components/ContentCard/ContentCard.tsx
@@ -32,8 +32,6 @@ export type ContentCardProps = {
   date?: string;
   /** Sets the image position on the card */
   imagePosition?: ImagePosition;
-  /** Sets an icon to be rendered over the image */
-  iconUrl?: string;
   /** Sets the background color according with the type */
   background?: Background;
   /** Sets the type of the clickable element */
@@ -50,6 +48,16 @@ export type ContentCardProps = {
   as?: React.ComponentProps<typeof Box.div>["as"];
   /** Overwrites default maxWidth */
   maxWidth?: BoxProps["maxW"];
+  /** Overwrites default h (height) */
+  h?: BoxProps["h"];
+  /** Overwrites default w (width) */
+  w?: BoxProps["w"];
+  /** Element to be rendered on the top left corner of the card */
+  topLeftElement?: ReactNode;
+  /** Element to be rendered on the top right corner of the card */
+  topRightElement?: ReactNode;
+  /** Applies reduced opacity to the card */
+  hasLowOpacity?: boolean;
 };
 
 const backgroundColor: Record<Background, BoxProps["backgroundColor"]> = {
@@ -69,12 +77,16 @@ export const ContentCard = ({
   text,
   tag,
   date,
-  iconUrl,
+  topLeftElement,
   imagePosition = "right",
   background = "default",
   target,
   as = "div",
   maxWidth,
+  h,
+  w,
+  topRightElement,
+  hasLowOpacity,
 }: ContentCardProps): JSX.Element => {
   const isImageOnTop = imagePosition === "top";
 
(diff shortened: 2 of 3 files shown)


.changeset/cold-news-glow.md
```diff
@@ -0,0 +1,14 @@
+---
+"@localyze-pluto/components": major
+---
+
+[ContentCard]
+
+> Breaking change: `iconUrl` prop is now `icon` and accepts a ReactNode
```
padulafacundo, line 7: Considering that it accepts ReactNode (which could be anything, not just an icon or an image), and that you are also adding topRightElement, why not call this topLeftElement?

.changeset/cold-news-glow.md
```diff
@@ -0,0 +1,14 @@
...
+
+Minor changes:
+
+- Set Icon wrapper border radius to 8px (shows now like a square)
+- Added new `topRightElement` prop to allow add element in the top right corner of the card
+- Added new `lowOpacity` prop to set the opacity of the card content without include the `topRightElement`
```
padulafacundo, line 13: nit: why not pass `opacity` (as a number) instead of lowOpacity ? It feels like we are not adding a prop but rather using one of the default css props

padulafacundo, line 13: Not setting the opacity on the topRightElement feels super specific, like something that should not be part of the DS, especially considering that the topRightElement could be anything... like, why is it only the topRIghtElement and not the icon too...

I don't have an alternative, I'm just thinking out loud here

.changeset/cold-news-glow.md
apvale: Because it's only an icon for the library, for Figma. We don't want to have a button now, for example, and I didn't want to restrict the type of the icon to string or Images from NextJS because for me it would be too specific.

.changeset/cold-news-glow.md
apvale: I’m not sure I understand what you mean. You mean the prop name? Do you think renaming it to `hasLowOpacity` would be better? 
I didn’t add it as a number since it’s a specific opacity value. In our case at Localyze, we want to apply a lower opacity (0.6) when the partner is disabled. I avoided using disabled as a prop because the card remains interactive.

apvale: Yeah, I get it. This was meant for our scenario on the partners page. We want to display the card with low opacity but still show the badge. 🤷🏻‍♀️

padulafacundo: I meant a number, yes. I know it's a specific number but right now it feels like we have so many specific props for this card. Anyway, not a blocker

.changeset/cold-news-glow.md
padulafacundo: I agree it would be too specific, but the name of the prop doesn't really match what the prop is for. Even if we don't want to have a button now, it is possible to do it. And considering that this component lives in pluto (DS) and not in the app, I would argue it doesn't really matter "how" we want to use it, but rather what the component is capable of

.changeset/cold-news-glow.md
padulafacundo: Yeah...

.changeset/cold-news-glow.md
apvale: Changed this here https://github.com/Localitos/pluto/pull/1752/files#diff-f7fa8c6df160defa8d1aad22a1a3e5af7da525824cca21b138dee5363ceedbd9R56

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -114,144 +126,168 @@ export const ContentCard = ({
...
+        display="flex"
+        flexDirection={
+          isImageOnTop ? "column-reverse" : { _: "column-reverse", md: "unset" }
+        }
+        fontFamily="fontFamilyNotoSans"
+        h={h}
+        justifyContent="start"
+        maxH={maxHeight[imagePosition]}
+        maxW={maxWidth || maxWidthDefault[imagePosition]}
+        opacity={lowOpacity ? "0.6" : "1"}
```
rssilva, line 156:
```suggestion
        opacity={lowOpacity ? 0.6 : 1}
```

rssilva (approved):
looks good! I have small suggestions but feel free to ignore it

packages/components/src/components/ContentCard/ContentCard.tsx
```diff
@@ -50,6 +48,16 @@ export type ContentCardProps = {
...
+  /** Overwrites default h (height) */
+  h?: BoxProps["h"];
+  /** Overwrites default w (width) */
+  w?: BoxProps["w"];
+  /** Element to be rendered on the top left corner of the card */
+  topLeftElement?: ReactNode;
+  /** Element to be rendered on the top right corner of the card */
+  topRightElement?: ReactNode;
+  /** Applies reduced opacity to the card */
+  lowOpacity?: boolean;
```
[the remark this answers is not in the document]
apvale: Changed it here https://github.com/Localitos/pluto/pull/1752/files#diff-f7fa8c6df160defa8d1aad22a1a3e5af7da525824cca21b138dee5363ceedbd9R60

packages/components/src/components/ContentCard/ContentCard.tsx
apvale: Changed it here https://github.com/Localitos/pluto/pull/1752/files#diff-f7fa8c6df160defa8d1aad22a1a3e5af7da525824cca21b138dee5363ceedbd9R156

PR #1765, opened 2024-11-01
Update utility card to use emoji instead of image

franziiiini:
## Description of the change
Now we save the category specific emoji in the DB so we don't need to load any emoji-images anymore.

- update UtilityCard to use emoji/string instead of image

## Type of change

Screenshot 2024-11-01 at 14 17 00

### Development

### Code review

This should only be merged with the relevant change in the Localitos/Backend (WIP)

### DIFF
--- a/packages/components/src/components/UtilityCard/UtilityCard.tsx
+++ b/packages/components/src/components/UtilityCard/UtilityCard.tsx
@@ -77,7 +77,7 @@ export const UtilityCard: React.FC<UtilityCardProps> = ({
         padding={{ _: "d4", md: "d6" }}
         w={{ _: "56px", md: "56px" }}
       >
-        <Box.div aria-hidden="true">{emoji}</Box.div>
+        <Box.span aria-hidden="true">{emoji}</Box.span>
       </Box.div>
 
       <Box.div
--- a/packages/components/src/components/UtilityCard/UtilityCard.tsx
+++ b/packages/components/src/components/UtilityCard/UtilityCard.tsx
@@ -77,7 +77,7 @@ export const UtilityCard: React.FC<UtilityCardProps> = ({
         padding={{ _: "d4", md: "d6" }}
         w={{ _: "56px", md: "56px" }}
       >
-        {emoji}
+        <Box.div aria-hidden="true">{emoji}</Box.div>
       </Box.div>
 
       <Box.div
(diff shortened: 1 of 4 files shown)


packages/components/src/components/UtilityCard/UtilityCard.tsx
```diff
@@ -9,8 +9,8 @@ export enum InteractiveElementTypeUtilityCard {
 }
 
 export type UtilityCardProps = {
-  /** Sets the card image source */
-  imageSrc: string;
+  /** Sets the category emoji */
```
apvale, line 12: I wouldn't add the `category` here because this component needs to be generic. So, we could say: `Set an emoji for the card`. What do you think?

packages/components/src/components/UtilityCard/UtilityCard.tsx
```diff
@@ -77,7 +77,7 @@ export const UtilityCard: React.FC<UtilityCardProps> = ({
         padding={{ _: "d4", md: "d6" }}
         w={{ _: "56px", md: "56px" }}
       >
-        <Box.img alt="" aria-hidden h="24px" src={imageSrc} w="24px" />
+        <Text.p>{emoji}</Text.p>
```
apvale, line 80: I think we don't even need a paragraph here. Having only the emoji string should be enough.

apvale (approved):

packages/components/src/components/UtilityCard/UtilityCard.tsx
```diff
@@ -77,7 +77,7 @@ export const UtilityCard: React.FC<UtilityCardProps> = ({
         padding={{ _: "d4", md: "d6" }}
         w={{ _: "56px", md: "56px" }}
       >
-        <Box.img alt="" aria-hidden h="24px" src={imageSrc} w="24px" />
+        {emoji}
```
padulafacundo, line 80: can we add `aria-hidden` to the wrapping div? the image used to have this attribute

packages/components/src/components/UtilityCard/UtilityCard.tsx
```diff
@@ -77,7 +77,7 @@ export const UtilityCard: React.FC<UtilityCardProps> = ({
         padding={{ _: "d4", md: "d6" }}
         w={{ _: "56px", md: "56px" }}
       >
-        <Box.img alt="" aria-hidden h="24px" src={imageSrc} w="24px" />
+        <Box.div aria-hidden="true">{emoji}</Box.div>
```
apvale, line 80: I would change it from `Box.div` to `Box.span` because the emoji acts like a simple text.

PR #1777, opened 2024-11-15
feat: use proper states and make UtilityCard more generic

apvale:
## Description of the change

Breaking Change:

- Removed the `interactiveElementType` prop in order to add two new states properties `hoverable` and `clickable`
- Renamed `categoryTag` to `subTitle` to make it more generic
- `onClick` prop is only required when using `clickable` as true

Minor:

- Added `hoverable` and `clickable` to have more control of state of the card
- Added `emojiWrapperSize` prop with options `small` and `large` so it's adaptable to different use cases
- Added `indicator` to add a helper element to the top right of the card

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/.changeset/green-keys-marry.md
+++ b/.changeset/green-keys-marry.md
@@ -1,0 +1,17 @@
+---
+"@localyze-pluto/components": major
+---
+
+UtilityCard:
+
+Breaking Change:
+
+- Removed the `interactiveElementType` prop in order to add two new states properties `hoverable` and `clickable`
+- Renamed `categoryTag` to `subTitle` to make it more generic
+- `onClick` prop is only required when using `clickable` as true
+
+Minor:
+
+- Added `hoverable` and `clickable` to have more control of state of the card
+- Added `emojiWrapperSize` prop with options `small` and `large` so it's adaptable to different use cases
+- Added `indicator` to add a helper element to the top right of the card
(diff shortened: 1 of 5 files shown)


packages/components/src/components/UtilityCard/UtilityCard.stories.tsx
```diff
@@ -32,7 +33,11 @@ Default.args = {
 export const CardWithBadge: Story = Template.bind({});
```
franziiiini, line 33: Maybe we can add CardWithContent here as well?

franziiiini (commented):
Looks good to me. Might be worthwhile to add a Story template. What do you think?

.changeset/green-keys-marry.md
```diff
@@ -0,0 +1,17 @@
...
+
+UtilityCard:
+
+Breaking Change:
+
+- Removed the `interactiveElementType` prop in order to add two new states properties `hoverable` and `clickable`
+- Renamed `categoryTag` to `subTitle` to make it more generic
+- `onClick` prop is only required when using `clickable` as true
+
+Minor:
```
padulafacundo, line 13: is it fine that this is a major update but we include minor updates?

apvale: From what we have [here](https://semver.org/#spec-item-8), I would say it's fine. Do you see any specific issue with it?

PR #1778, opened 2024-11-19
feat: make improvements to Menu component

apvale:
## Description of the change

While working on the Partner Preferences page and adding the menu `Set preferences` there, I noticed that we had some issues with the Menu component and decided to make some improvements. Changelog:

* Breaking change: Removed the unnecessary div wrapper from the Menu component; props are now passed directly to the Menu.
* Improved focus behavior for menu items by adding a background color to the focused item
* Added a new customStore prop to enable custom store settings and support different behaviors, such as placement
* Set a default gutter value of 3 for the Menu component to align with the Figma design
* Introduced a menuButtonProps prop to enable passing props to the default menu button

This component was limited to its own store. With this change, we can now customize things like the gutter, a custom store, and menu popover position. We can also import the store directly from Pluto instead of ariakit.

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Menu/Menu.stories.tsx
+++ b/packages/components/src/components/Menu/Menu.stories.tsx
@@ -6,6 +6,7 @@ import { Button } from "../Button";
 import { Box } from "../../primitives/Box";
 import { Icon } from "../Icon";
 import { Menu, MenuItemProps } from "./Menu";
+import { useMenuStore } from "./index";
 
 const meta: Meta<typeof Menu> = {
   title: "Components/Menu",
@@ -15,7 +16,7 @@ const meta: Meta<typeof Menu> = {
 export default meta;
 type Story = StoryObj<typeof Menu>;
 
-const DefaultMenu = (): JSX.Element => {
+const DefaultMenu = (): React.JSX.Element => {
   const items: MenuItemProps[] = [
     {
       onClick: noop,
@@ -35,7 +36,7 @@ const DefaultMenu = (): JSX.Element => {
 };
 
 export const Default: Story = {
-  render: (): JSX.Element => <DefaultMenu />,
+  render: (): React.JSX.Element => <DefaultMenu />,
 };
 
 Default.parameters = {
@@ -45,7 +46,7 @@ Default.parameters = {
   },
 };
 
-const CustomButtonMenu = (): JSX.Element => {
+const CustomButtonMenu = (): React.JSX.Element => {
   const items: MenuItemProps[] = [
     {
       onClick: noop,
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Menu/Menu.stories.tsx
```diff
@@ -61,13 +62,44 @@ const CustomButtonMenu = (): JSX.Element => {
...
+        label: "Option 2",
+      },
+    ];
+
+    return (
+      <Box.div minHeight="100px">
+        <Menu
+          customStore={store}
+          items={items}
+          menuButtonProps={{
```
apvale, line 96: I didn't want to create another story only to show the change in the menuButtonProps.

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -40,24 +45,42 @@ const FullWidthButton = ({ ...props }) => (
...
+  ) => {
+    const store = useMenuStore({
+      store: customStore,
+    });
 
     const button = menuButton || VerticalEllipsisButton;
 
     return (
-      <Box.div {...props}>
+      <>
```
apvale, line 66: When removing this, we will need to fix the zIndex in the classic app, for example, because the prop won't be accepted there anymore. That's why I created the breaking change.

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -71,41 +94,48 @@ const Menu = React.forwardRef<HTMLButtonElement, BoxProps & MenuProps>(
...
-                    justifyContent="flex-start"
-                    onClick={onClick}
-                    padding="d0"
-                    px="d4"
-                    py="d2"
-                    variant="ghost"
-                    w="100%"
+              <MenuItem
+                key={i}
+                render={
```
apvale, line 99: The render was used to avoid the wrong outline added to the menu item:

https://github.com/user-attachments/assets/bf1964f2-bbf5-43ab-a438-29985e35d9d0

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -15,14 +16,18 @@ export type MenuItemProps = {
   disabled?: boolean;
 };
 
-export type MenuProps = {
+export interface MenuProps extends Omit<AriakitMenuProps, "store"> {
```
rssilva, line 19: maybe if we want to avoid using an interface here just because of the `extends`:

```suggestion
export type MenuProps = Omit<AriakitMenuProps, "store"> & {
```

rssilva (approved):
💯 

Left one suggestion but feel free to ignore

PR #1788, opened 2024-12-30
add new Combobox component

apvale:
## Description of the change

Introduces new `Combobox` component:

- Combines a text input and a popover to allow users to filter from long lists of options
- Can be used as a standalone component or composed with other components such as `ComboboxInput`, `ComboboxPopover`, `ComboboxItem`, and `ComboboxProvider`
- Supports generic type for the options list
- Provides a prop `filterItem` prop to be used as the filter function when receiving a function. 

## Testing the change

## Type of change

### Development

### Code review

### DIFF
--- a/packages/components/src/components/Combobox/Combobox.stories.tsx
+++ b/packages/components/src/components/Combobox/Combobox.stories.tsx
@@ -1,0 +1,246 @@
+import type { Meta, StoryFn } from "@storybook/react";
+import React, { useEffect, useState } from "react";
+import toLower from "lodash/toLower";
+import includes from "lodash/includes";
+import map from "lodash/map";
+import { useComboboxStore } from "@ariakit/react";
+import { Controller, SubmitHandler, useForm } from "react-hook-form";
+import find from "lodash/find";
+import { Box } from "../../primitives/Box";
+import { Label } from "../Label";
+import { HelpText } from "../HelpText";
+import { Button } from "../Button";
+import { Combobox, ComboboxItemProps, ComboboxProps } from "./Combobox";
+import { ComboboxInput } from "./ComboboxInput/ComboboxInput";
+import { ComboboxPopover } from "./ComboboxPopover/ComboboxPopover";
+import { ComboboxItem } from "./ComboboxItem/ComboboxItem";
+
+export default {
+  component: Combobox,
+  title: "Components/Combobox",
+} as Meta<typeof Combobox>;
+
+const items: ComboboxItemProps[] = [
+  { value: "1", label: "Option One" },
+  { value: "2", label: "Option Two" },
+  { value: "Option Three", label: "Option Three" },
+  { value: "Option Four", label: "Option Four" },
+];
+
+const Template: StoryFn<ComboboxProps> = ({ children, ...props }) => (
+  <Combobox {...props}>{children}</Combobox>
+);
+
+export const Default = Template.bind({});
+Default.args = {
+  items,
+  placeholder: "Select an option",
+  noResultsMessage: "No results found",
+  filterItem: true,
+  hasError: false,
+};
+Default.parameters = {
+  docs: {
+    description: {
+      story: "The combobox will filter the items based on the input value.",
+    },
+  },
+};
+
+export const WithOnChange = (): React.JSX.Element => (
+  <Combobox
+    items={items}
+    onChangeInput={(event) => {
+      // eslint-disable-next-line no-console
+      console.log(event.target.value);
+    }}
+    placeholder="Select an option"
+  />
+);
+WithOnChange.parameters = {
+  docs: {
+    description: {
+      story:
+        "The combobox will call the `onChangeInput` function when the input value changes.",
+    },
+  },
+};
+
+export const WithOnSelectItem = (): React.JSX.Element => (
+  <Combobox
+    items={items}
+    onSelectItem={(value) => {
+      // eslint-disable-next-line no-console
+      console.log(value);
+    }}
+    placeholder="Select an option"
+  />
+);
+WithOnSelectItem.parameters = {
+  docs: {
+    description: {
+      story:
+        "The combobox will call the `onSelectItem` function when an option is selected.",
+    },
+  },
+};
+
+export const WithLongItem = (): React.JSX.Element => {
+  const longItem = {
+    value:
+      "Option Five which is a really long item and should overflow to the next line",
+    label:
+      "Option Five which is a really long item and should overflow to the next line",
+  };
+
+  return (
+    <Box.div w="250px">
+      <Label htmlFor="select-option">Select an option</Label>
+      <Combobox id="select-option" items={[...items, longItem]} name="select" />
+    </Box.div>
+  );
+};
+
+type CustomItem = {
+  value: string;
+  label: string;
+  disabled?: boolean;
+};
+
+export const WithCustomFilterItem = (): React.JSX.Element => {
+  const customItems: CustomItem[] = [
+    ...items,
+    { value: "New option 1", label: "New option 1", disabled: true },
+    { value: "New option 2", label: "New option 2", disabled: true },
+  ];
+
+  const customFilter = (inputValue: string, item: CustomItem): boolean =>
+    includes(toLower(item.label), toLower(inputValue)) && !item.disabled;
+
+  return <Combobox filterItem={customFilter} items={customItems} />;
+};
+WithCustomFilterItem.parameters = {
+  docs: {
+    description: {
+      story:
+        "The combobox will filter the items based on a given function " +
+        "that filters the items based on the input value and the item.",
+    },
+  },
+};
+
+export const WithReactHookForm = (): React.JSX.Element => {
+  interface FormInputs {
+    item: string;
+  }
+
+  const [selectedLabel, setSelectedLabel] = useState<string | undefined>("");
+  const { control, handleSubmit, clearErrors, formState } = useForm({
+    defaultValues: {
+      item: "",
+    },
+  });
+
+  useEffect(() => {
+    clearErrors();
+  }, [clearErrors, formState.isDirty]);
+
+  // eslint-disable-next-line unicorn/consistent-function-scoping
+  const onSubmit: SubmitHandler<FormInputs> = (data) =>
+    alert(JSON.stringify(data, null, 2));
+
+  return (
+    <Box.form noValidate onSubmit={handleSubmit(onSubmit)}>
+      <Controller
+        control={control}
+        name="item"
+        render={({ field: { onChange, ...field }, fieldState }) => (
+          <>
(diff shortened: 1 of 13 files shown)


packages/components/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,148 @@
...
+              {item.label}
+            </ComboboxItem>
+          ))
+        )}
+      </ComboboxPopover>
+    </ComboboxProvider>
+  );
+};
+
+const Combobox = forwardRef(ComboboxInner) as <T extends ComboboxItemProps>(
```
apvale, line 143: Initially, when I opened the PR, I forgot to pass the type `T` so we could use it in the apps. When I tried to fix it here, I had some issues when using forwardRef. I managed to fix it by using the following as an example: https://gist.github.com/IrvingArmenta/a6d7fc76ed538697ad18b7f074accdde.

packages/components/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+};
+Default.parameters = {
+  docs: {
+    description: {
+      story: "The combobox will filter the items based on the input value.",
+    },
+  },
+};
+
+export const WithoutFilterItem = Template.bind({});
```
padulafacundo, line 48: when would we not filter?

packages/components/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+  },
+};
+
+export const WithOnChange = Template.bind({});
+WithOnChange.args = {
+  ...Default.args,
+  onChangeInput: (event) => {
+    // eslint-disable-next-line no-console
+    console.log(event.target.value);
+  },
```
padulafacundo, line 67: this works ok but when I click `show code` it shows the onChangeInput function as a void function.. do you know why that is? (same for onSelectItem etc)

image

packages/components/src/components/Combobox/ComboboxPopover/ComboboxPopover.tsx
```diff
@@ -0,0 +1,41 @@
...
+      outlineColor={{ focus: "colorBorderPrimary" }}
+      outlineStyle={{ focus: "solid" }}
+      outlineWidth={{ focus: "borderWidth10" }}
+      overflow="auto"
+      overscrollBehavior="contain"
+      portal
+      ref={ref}
+      sameWidth
+      unmountOnHide
+      zIndex="zIndex50"
```
padulafacundo, line 33: why 50 ?

packages/components/src/components/Combobox/ComboboxPopover/ComboboxPopover.tsx
```diff
@@ -0,0 +1,41 @@
...
+      backgroundColor="bgPrimary"
+      borderRadius="borderRadius20"
+      boxShadow="shadowStrong"
+      display="flex"
+      flexDirection="column"
+      gutter={3}
+      maxHeight="250px"
+      outlineColor={{ focus: "colorBorderPrimary" }}
+      outlineStyle={{ focus: "solid" }}
+      outlineWidth={{ focus: "borderWidth10" }}
```
padulafacundo, line 26: I think `focus` is not working as expected here, I cannot see the blue outline. I had to use the devtools and simulate the focus on the specific html element to see it

packages/components/src/components/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { userEvent } from "@testing-library/user-event";
+import { Combobox } from "./Combobox";
+
+const SELECT_AN_OPTION_LABEL = "Select an option";
+
+describe("<Combobox />", () => {
+  const items = [
+    { label: "Item 1", value: "Item 1" },
+    { label: "Item 2", value: "Item 2" },
+    { label: "Item 3", value: "Item 3" },
```
padulafacundo, line 12: wouldnt it be better to test this with values that are different to the labels? to make sure that we dont mix them up

packages/components/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,148 @@
...
+        }}
+        placeholder={placeholder}
+        ref={ref}
+        store={store}
+        textOverflow="ellipsis"
+        {...props}
+      />
+      <ComboboxPopover sameWidth>
+        {filteredItems.length === 0 ? (
+          <ComboboxItem>{noResultsMessage}</ComboboxItem>
```
padulafacundo, line 124: isn't it a little bit confusing that we show this message but we still allow users to enter values that are not on the list?

packages/components/src/components/Combobox/Combobox.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+              Select an option
+            </Label>
+            <Combobox
+              {...field}
+              aria-describedby="select-option-help-text"
+              hasError={fieldState.invalid}
+              id="select-option"
+              items={items}
+              onChangeInput={(event) => field.onChange(event.target.value)}
+              onSelectItem={(value) => field.onChange(value)}
```
padulafacundo, line 173: when selecting an item the label is replaced with the value of the option. Example:
1) type "two"
2) select option "Option Two"
3) input now shows "2" instead of "Option Two"

packages/components/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,148 @@
...
+      ? (matchSorter(items as ComboboxItemProps[], searchValue) as T[])
+      : items;
+  }
+  if (isFunction(filterItem)) {
+    return filter(items, (item) => filterItem(searchValue, item));
+  }
+  return items;
+};
+
+// eslint-disable-next-line jsdoc/require-param
```
rssilva, line 71: we should add `jsdoc/require-returns` here too

packages/components/src/components/Combobox/Combobox.tsx
apvale: Why do you think it's confusing? What do you think it should happen? 

For me, it's clearer to show a message when there are no results for an entry. I think it opens the possibility to add new values when we don't have them. Also, we already have this behavior in our current "Autocomplete" component and I wanted to keep it. Besides that, this is only for the `Combobox` component. The consumer is still able to compose a new component and avoid using the no results message.

packages/components/src/components/Combobox/Combobox.stories.tsx
apvale: I couldn't find a reason for why it's showing like that, but I discovered this https://storybook.js.org/docs/6/writing-docs/doc-blocks/doc-block-source and changed the stories properly to be able to show the callback. Changed here https://github.com/Localitos/pluto/pull/1788/commits/ee1b9df47f822dda26d23cf9f418d26bfb003584

packages/components/src/components/Combobox/ComboboxPopover/ComboboxPopover.tsx
apvale: I used the same styles from talent-web, but now I see we don't need them. The focused element should be an option when navigating down from the input, not the popover.

packages/components/src/components/Combobox/ComboboxPopover/ComboboxPopover.tsx
apvale: > why 50 ? 
Sorry, I just used what we had on the talent-web. 

Following what Rafa suggested, I added a `zIndex` prop with `zIndex50` as default.

packages/components/src/components/Combobox/Combobox.tsx
apvale: Done [here](https://github.com/Localitos/pluto/pull/1788/files#diff-395d307cf2a5c77a478d9c4ed84231437e5ee35ebc7e642c119997cdafc3c7ffR72)

packages/components/src/components/Combobox/Combobox.test.tsx
apvale: Good idea! Done [here](https://github.com/Localitos/pluto/pull/1788/files#diff-054dc0d4d697e47e94abef3f9fd8ec006ccd0addf0a3996f5bb1b6609faf6b8dR10)

packages/components/src/components/Combobox/Combobox.stories.tsx
apvale: I couldn't remember a scenario where we would not filter. I researched some design systems and a few have the "default" without the filter, that's why I added it here. But IMO, it's better to use a "Combobox" that lets you search and match items. Either way, I kept the `filterItem` prop in case we need it but removed the `WithoutFilterItem` story as it doesn’t seem necessary.

packages/components/src/components/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    const options = screen.getAllByRole("option");
+    expect(options).toHaveLength(3);
+
+    await user.type(combobox, "Item 4");
+
+    const noResults = screen.getByRole("option", { name: "No results found" });
+    expect(noResults).toBeInTheDocument();
+  });
+
+  it("selects an item", async () => {
```
rssilva, line 69: nit: I believe that this test can be "merged" with 'calls onSelectItem when selecting an item'. What do you think? They are similar almost being redundant

packages/components/src/components/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { ComboboxItem } from "./ComboboxItem/ComboboxItem";
+import { ComboboxProvider } from "./ComboboxProvider/ComboboxProvider";
+
+export type ComboboxItemProps = {
+  label: string;
+  value: string;
+  disabled?: boolean;
+};
+
+export type ComboboxProps<T extends ComboboxItemProps = ComboboxItemProps> =
```
rssilva, line 24: Because `{...props}` are being passed to the input we could pass tokens as below and that will work

``` typescript
<Combobox
      backgroundColor="bgError"
      borderColor="bgError"
      borderStyle="borderStyleSolid"
      borderWidth="borderWidth30"
      filterItem={customFilter}
      items={customItems}
      padding="d5"
/>
```

so the type here would need to be

```
export type ComboboxProps<T extends ComboboxItemProps = ComboboxItemProps> =
  BoxProps &
    Omit<AriakitComboboxProps, "store"> & {

```

rssilva (approved):
Nice! Just one small suggestion and a fix to the Props type

packages/components/src/components/Combobox/Combobox.test.tsx
apvale: Yes, I agree with you. Already changed in the last commit. 🙏🏻

PR #1817, opened 2025-02-11
chore: remove component specific tokens in favor of more general

rssilva:
## Description of the change

- Remove component specific tokens in favor of global ones

## Type of change


packages/theme/src/__tests__/__snapshots__/theme.test.tsx.snap
```diff
@@ -51,7 +47,15 @@ exports[`Theme matches the default theme 1`] = `
     "bgError": "#FEF2F2",
     "bgInfo": "#EBEDFF",
     "bgPrimary": "#FFFFFF",
+    "bgPrimaryActive": "#F8FAFC",
```
apvale, line 50: I noticed that besides these ones, we also have a few tokens related to `bg-branded` [here](https://www.figma.com/design/uDEjU77uIrQ5T1csn8IvIJ/Pluto-Tokens?node-id=2755-457&m=dev). Are you going to add them to this PR as well?

rssilva: I prefer doing that in a different one because all the cache crazyness on github actions

